PR #3, opened 2021-12-30
[SWUI-82] Check if all icons in the SDK are configurable

DaemonLoki:
### 🔗 Issue Link
[_Jira or Github issue link, if applicable._](https://stream-io.atlassian.net/browse/SWUI-82)

### 🎯 Goal

We want all Images to configurable so the source should always be in the `Images.swift` file.

### 🛠 Implementation

All images that were directly created in the components and views have been changed to use the injected `images` reference.

### 🧪 Testing

Possible to test if you can replace all the images used.

### 🎨 Changes

-

### ☑️ Checklist


martinmitrevski (changes requested):
Code-wise looks good. However, while testing the app I've noticed several differences with the previous images. For example, the video icon in the gallery picker is too big, the eye is not grey, share and xmark look strange. 
My assumption is that this happens in places where we directly replace Image(systemName:) with Image(uiImage:) - it seems the modifiers are not applied. I guess in those cases we can expose directly Image (or you might have a better idea?).
Please compare with the base branch to make sure there are no breaking UI changes.

DaemonLoki:
I stayed with the way of using `UIImage` for consistency reasons.

However, I re-evaluated each of the used icons and adjusted the UI to be the same as before. This required some hard-coding of sizes where we did use things like the `.font` modifiers for SF Symbols. I think it makes sense to omit those when users are technically able to replace these icons with any custom images they like.

My approach was to achieve pixel-perfect replicas of the current design by using `ZStack`s with the previous implementation and matching the new implementation with that. 

Let me know if there's still things to be improved, happy to take a look again!

Sources/StreamChatSwiftUI/ChatChannel/Gallery/ShareButtonView.swift
```diff
@@ -16,8 +17,11 @@ struct ShareButtonView: View {
         Button(action: {
             self.isSharePresented = true
         }, label: {
-            Image(systemName: "square.and.arrow.up")
-                .font(fonts.bodyBold)
+            Image(uiImage: images.share)
+                .renderingMode(.template)
```
martinmitrevski, line 21: Since these three modifiers are used a lot, it might make sense to group them in one modifier and apply that one to the image.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/VideoAttachmentView.swift
```diff
@@ -86,10 +88,12 @@ public struct VideoAttachmentView: View {
...
                 } label: {
-                    Image(systemName: "play.fill")
-                        .font(.system(size: 30))
+                    Image(uiImage: images.playFilled)
+                        .renderingMode(.template)
+                        .resizable()
+                        .scaledToFit()
+                        .frame(width: 24)
                         .foregroundColor(.white)
-                        .padding(.all, 32)
```
martinmitrevski, line 92: This was used to increase the tappable area of the play control. Is there any specific reason why it was removed?

Sources/StreamChatSwiftUI/ChatChannel/MessageList/DeletedMessageView.swift
```diff
@@ -28,14 +29,18 @@ public struct DeletedMessageView: View {
                 .standardPadding()
                 .foregroundColor(Color(colors.textLowEmphasis))
                 .messageBubble(for: message, isFirst: isFirst)
+            
             HStack {
                 Spacer()
-                Image(systemName: "eye")
+                
+                Image(uiImage: images.eye)
```
martinmitrevski, line 36: The eye icon is still black, it should be grey like in the current version. I think we need to explicitly provide the color in this case.

Sources/StreamChatSwiftUI/ChatChannel/Gallery/ShareButtonView.swift
```diff
@@ -16,8 +17,11 @@ struct ShareButtonView: View {
         Button(action: {
             self.isSharePresented = true
         }, label: {
-            Image(systemName: "square.and.arrow.up")
-                .font(fonts.bodyBold)
+            Image(uiImage: images.share)
```
martinmitrevski, line 20: The share icon looks a bit different. It's not bold anymore, and it seems the image ratio is changed.

martinmitrevski (changes requested):
Additionally, we need to: 
- Check the remaining 2-3 views still using system image directly
- Change the PR to point to main (first please merge main to the branch)

Sources/StreamChatSwiftUI/ChatChannel/MessageList/VideoAttachmentView.swift
DaemonLoki: I was wondering if it fills any purpose but couldn't find it. That's why I removed it, which was a mistake. Reverted the change.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/DeletedMessageView.swift
DaemonLoki: Fixed through the usage of the custom modifier!

Sources/StreamChatSwiftUI/ChatChannel/Gallery/ShareButtonView.swift
DaemonLoki: Done. Needed to extend `Image` here as it could not be a regular custom modifier due to things like `.resizable` only being applicable to `Image` itself. However, I think it still fits best in the `Modifiers` file.

DaemonLoki:
> Check the remaining 2-3 views still using system image directly

The only remaining direct usages of `systemName` are either in tests (I think that's fine to directly use them here), mocks (also fine IMO), and the Demo App. Correct me if I'm wrong and I'm happy to change those.

martinmitrevski:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please remove period from end of commit subject line.<br />
c60393e6560971770a4bce9ec4cd5b969f1f9d22</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">There seems to be app changes but CHANGELOG wasn't modified.<br />
Please include an entry if the PR includes user-facing changes.<br />
You can find it at <a href="https://github.com/GetStream/stream-chat-swift/blob/main/CHANGELOG.md">CHANGELOG.md</a>.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

martinmitrevski (approved):
Looks good to me!

PR #63, opened 2022-05-10
MessageView: check for deletions before checking for customAttachment

AndrewSB:
might fix https://linear.app/reclip/issue/R-1005#comment-79078248, aka https://linear.app/reclip/issue/R-1104


martinmitrevski (approved):
Thanks @AndrewSB for the contribution!

PR #69, opened 2022-05-15
ChatChannelListView: remove vertical padding

AndrewSB:
without this, there's padding between the top view/ bottom sticky view in the channel list that a client app can't get rid of


martinmitrevski:
Hey @AndrewSB, thanks for creating the PR. This requires some further work and new snapshots generation, so we will handle this internally.

martinmitrevski:
Hey @AndrewSB, thanks for creating the PR. This requires some further work and new snapshots generation, so we will handle this internally.

PR #76, opened 2022-05-24
Ensure URL have secure scheme + Add view info on channel options

adolfogarza:
### 🔗 Issue Link
https://stream-io.atlassian.net/browse/SWUI-161
https://stream-io.atlassian.net/browse/SWUI-155

### 🎯 Goal

_Ensure URL has a secure scheme so it can be opened properly.

_Add View Info section to channel slide options._

### 🛠 Implementation

_URL has an internal extension that can be called by using `secureURL` which ensures that the URL has a safe scheme which entails that links can be opened even if the scheme is missing and that the app will point to the safe version of the sites._

_View info was added as a new action within the default actions, with the goal of presenting through a full screen sheet the `ChatChannelInfoView ` when tapped_

### ☑️ Checklist


Sources/StreamChatSwiftUI/ChatChannelList/MoreChannelActionsView.swift
```diff
@@ -91,6 +106,24 @@ public struct MoreChannelActionsView: View {
         .onTapGesture {
             onDismiss()
         }
+        .fullScreenCover(isPresented: $isPresented) {
+            if let fullScreenView = presentedView {
+                ZStack(alignment: .topTrailing) {
+                    fullScreenView
```
martinmitrevski, line 112: We still need the nav bar view here, since it displays some channel information. We can put the close on the left side, similarly to the GalleryView.

Sources/StreamChatSwiftUI/ChatChannelList/MoreChannelActionsView.swift
```diff
@@ -91,6 +106,24 @@ public struct MoreChannelActionsView: View {
...
+                    
+                    Button {
+                        isPresented = false
+                        presentedView = nil
+                    } label: {
+                        Image(uiImage: images.closeFilled)
+                            .customizable()
+                            .frame(height: 30)
```
martinmitrevski, line 120: The button is too big, maybe 24?

Sources/StreamChatSwiftUI/ChatChannelList/MoreChannelActionsView.swift
```diff
@@ -91,6 +106,24 @@ public struct MoreChannelActionsView: View {
...
+                            .padding(15)
```
martinmitrevski, line 121: We shouldn't use arbitrary numbers for paddings. If we have to, then the default ones should be applied (8,16, etc). In this case, only .padding() is enough, since it will apply the default 16 padding.

Sources/StreamChatSwiftUI/ChatChannelList/MoreChannelActionsView.swift
```diff
@@ -91,6 +106,24 @@ public struct MoreChannelActionsView: View {
...
+                            .foregroundColor(.black)
```
martinmitrevski, line 122: We shouldn't use .black as a color. This makes the icon not visible in dark mode. Use a color from our color pallete (e.g. text). And please check if everything is clearly visible in dark mode.

Sources/StreamChatSwiftUI/ChatChannelList/MoreChannelActionsView.swift
```diff
@@ -49,18 +51,31 @@ public struct MoreChannelActionsView: View {
...
                             .padding(.horizontal, -16)
                         
-                        Button {
-                            if action.confirmationPopup != nil {
-                                viewModel.alertAction = action
-                            } else {
-                                action.action()
+                        if let destination = action.navigationDestination {
+                            Button {
+                                presentedView = destination
```
martinmitrevski, line 56: Maybe we can make these properties dependent? For example, on the didSet of `presentedView` we could change the `isPresented` bool value. With the current approach, we have to always set the two values together.

StreamChatSwiftUITests/Tests/Utils/URLUtils_Tests.swift
```diff
@@ -0,0 +1,41 @@
+//
+// Copyright © 2022 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChat
+@testable import StreamChatSwiftUI
+import XCTest
+
+class URLUtils_Tests: XCTestCase {
+    func test_secureSchemeURL_withMissingScheme() {
```
martinmitrevski, line 10: We are missing a test for the positive scenario (https://).

Sources/StreamChatSwiftUI/ChatChannelList/MoreChannelActionsView.swift
```diff
@@ -91,6 +106,24 @@ public struct MoreChannelActionsView: View {
         .onTapGesture {
             onDismiss()
         }
+        .fullScreenCover(isPresented: $isPresented) {
+            if let fullScreenView = presentedView {
+                ZStack(alignment: .topTrailing) {
```
martinmitrevski, line 111: We need a snapshot test for this view too.

Sources/StreamChatSwiftUI/ChatChannelList/DefaultChannelActions.swift
```diff
@@ -201,6 +206,20 @@ extension ChannelAction {
         return leaveConversation
     }
     
+    private static func viewInfo(for channel: ChatChannel) -> ChannelAction {
```
martinmitrevski, line 209: Maybe write a test to verify the existence of this option?

martinmitrevski (changes requested):
Looks good 👍 I've added few remarks that we need to fix before merging.

Sources/StreamChatSwiftUI/ChatChannelList/DefaultChannelActions.swift
adolfogarza: This is private so when I tried it cannot retrieve such option.

martinmitrevski (approved):
Great work @adolfogarza 👏

PR #89, opened 2022-05-31
Add token class to dynamically determine when SwiftUI SDK is used

polqf:
Needed for this PR: https://github.com/GetStream/stream-chat-swift/pull/2035

### 🎯 Goal

We were trying to know the linked SDKs by using canImport. This only works for built/statically linked frameworks. That's why it was working in the Demo app, which uses SPM, but not in the integration apps when using CocoaPods and Carthage.

### 🛠 Implementation

We will dynamically look for the availability of token classes added to both UIKit and SwiftUI SDKs, named SDKIdentifier, to determine if those libraries are linked or not.


martinmitrevski (approved):
Nice one 👍

PR #93, opened 2022-06-01
Remove DEVELOPMENT_TEAM from main target

finestructure:
The DEVELOPMENT_TEAM can interfere when running `xcodebuild docbuild` without running `xcodebuild build` in between.

I _think_ removing this shouldn't interfere with anything else but I've only made the change on the main target to keep the changes minimal. I've also checked the project in the related repo https://github.com/GetStream/stream-chat-swift.git and there the code signing settings are "Manual" without a team set, learning me to believe this was an oversight in this repo.

See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1773#issuecomment-1143311278 for details.


martinmitrevski:
Hey @finestructure,

Thanks for raising the PR. We've already removed the development team, @jeroenleenarts was a bit faster. 😃 

Here's the PR for reference:
https://github.com/GetStream/stream-chat-swiftui/pull/92

Best,
Martin

PR #101, opened 2022-06-08
[SWUI-181] Configure CI for UI tests

alteral:
### 🔗 Issue Link
- https://stream-io.atlassian.net/browse/SWUI-181


.github/workflows/smoke-checks.yml
```diff
@@ -4,13 +4,18 @@ on:
   push:
     branches:
       - main
+      - develop
```
martinmitrevski, line 7: We still use only a `main` branch.

PR #102, opened 2022-06-08
MessageDisplayInfo: make struct initalizable from outside module

AndrewSB:
This was causing an issue in our app where the makeMessageContainerView
factory method wasn't being picked up


martinmitrevski (approved):
Thanks for spotting this, @AndrewSB 👍

PR #116, opened 2022-07-01
trivial: ParticipantInfo: make init public

AndrewSB:
hey! we needed to create this struct from within our target, i just went ahead and created a public initializer


martinmitrevski (approved):
Thanks, @AndrewSB 👍

PR #119, opened 2022-07-04
Scroll to bottom long message bugfix

adolfogarza:
### 🔗 Issue Link
_https://stream-io.atlassian.net/browse/SWUI-177_

### 🎯 Goal

_When pressing the scroll to bottom button, the messages should be scrolled all the way down, even if the messages are very long._


martinmitrevski (commented):
With this, when you open the screen, there's a slight movement of the message list, which doesn't look good. Might be worth checking the paddings if we go with this approach.

PR #125, opened 2022-07-09
MessageTextView: make init public

AndrewSB:
wanted to reuse this instead of copy-pasting into our target


martinmitrevski (approved):
thanks @AndrewSB 👍

PR #132, opened 2022-07-13
add showLastInGroup property

saidReclip:
Hey, just added similar to `showsAllInfo` property called `showsLastInGroup`.
Goal is to detect which message is first or last in group of messages.


martinmitrevski:
Heu @saidReclip,

Thanks for creating the PR. However, we can't merge this, since it's a breaking change. We can only change method signatures in the `ViewFactory` on major version releases. We will add this when we do such major version release (no timeframe atm).

Alternatively, you can use `func makeLastInGroupHeaderView(for message: ChatMessage) -> some View`. Basically, it allows you to provide a view in the last message in the group (the oldest that was sent in a group), so it should cover your case.

Hope this helps,
Martin

Issue #122, opened 2022-07-06
Getting booted out of channel after editing channel name


AndrewSB:
## What did you do?
1. Navigate into a channel that you are a member in (not the owner, someone else had to create the channel)2. 
2. Rename the channel 

## What did you expect to happen?
To be in the channel, with the new name

## What happened instead?
The channel is successfully renamed, but the channel is immediately dismissed and you're returned to the inbox.
 
We get an error log:
"SwiftUI encountered an issue when pushing a NavigationLink. Please file a bug."

and based on my research, this might be a possible solution: https://stackoverflow.com/a/67626758
and an edit you'd make around https://github.com/GetStream/stream-chat-swiftui/blob/main/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift#L142

## GetStream Environment
**GetStream Chat version:** 4.17.0
**GetStream Chat frameworks:** StreamChatSwiftUI `main`
**iOS version:** 15.4
**Swift version:** 5
**Xcode version:** 13.4.1
**Device:** simulators, and iPhone 12

martinmitrevski:
Hey @AndrewSB,

Thanks for reporting the issue. We will analyse it and provide an update soon.

Best,
Martin

adolfogarza:
Hi @AndrewSB, do you know if the simulator is the one that creates the channel, or is the device the user who creates the channel?

AndrewSB:
within our project, doesn't seem to repro within the TestFlight of your demo app.
its possible that its coming from simulator created channels, i'll post back with more details once we have them

PR #133, opened 2022-07-13
move navigation link out of the ForEach loop

saidReclip:
This PR is fixing booting out from the channel issue. It is hard to reproduce it on default stream test due to different chat members rights. 

Possible STR's: 
- Be In a group you did not create
- Rename group to something else
- Go back into group and remove the name of the group leaving an empty string
- Save and go back to the group
- Open Settings

The idea is to move `NavigationLInk` out of `ForEach` loop on ChatChannelList.swift:152.
This is suggested solution. Maybe there is better one. 

How I came to this solution: [stackoverflow question](https://stackoverflow.com/questions/66017531/swiftui-navigationlink-bug-swiftui-encountered-an-issue-when-pushing-anavigatio/67626758#67626758)

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
@@ -137,6 +137,17 @@ struct ChannelsLazyVStack<Factory: ViewFactory>: View {
     }
     
     public var body: some View {
+        if let channel = selectedChannel?.channel {
+            NavigationLink(
+                tag: channel.channelSelectionInfo,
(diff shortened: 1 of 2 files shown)


martinmitrevski:
Hey @saidReclip,

Thanks for creating the PR. However, this change breaks our navigation transition (when opening a new channel), so we can't merge this in.

@adolfogarza is analyzing the ticket https://github.com/GetStream/stream-chat-swiftui/issues/122 connected with this issue.
 
Unfortunately, he's on sick leave at the moment, but as soon as he's back, he will tackle this again.

Best,
Martin

PR #134, opened 2022-07-13
Make MessageListConfig properties public

AndrewSB:
needed these to be public to grab the transition options


martinmitrevski (approved):
Thanks @AndrewSB for the contribution

PR #135, opened 2022-07-15
InputTextView: make becomeFirstResponder notification name public

AndrewSB:
I wasn't sure what the best way to solve this was, but i needed a way to trigger the input text to become the first responder. any ideas on a better public API


martinmitrevski (commented):
Hey @AndrewSB, as far as I remember, you are using your custom `InputTextView`, therefore is this really needed?

If yes, then I would suggest changing the constant name to sth like `getStreamFirstResponderNotification`.

AndrewSB:
yup, we use our own; you're right

PR #138, opened 2022-07-18
InputTextView: make becomeFirstResponder notification name public

AndrewSB:
related to https://github.com/GetStream/stream-chat-swiftui/pull/135

we've removed our custom `InputTextView`, using the default one instead


martinmitrevski (approved):
Thanks @AndrewSB

PR #140, opened 2022-07-19
MessageView: make text edge insets injectable

AndrewSB:
For messages that have a quoted reply above the message text, we wanted to reduce the padding


martinmitrevski:
Hey @AndrewSB,

Thanks for creating the PR. We want to keep the `StandardPaddingModifier` internal, in case of changes in the future. Also, there's no need to use edge insets in this case. 

I've added the same feature in a slightly different way, so you can now use `horizontalPadding` and `verticalPadding` in the init of the `MessageTextView`.

You can find more details here: https://github.com/GetStream/stream-chat-swiftui/commit/9ffda5834beb907978084aa1a0454906fd4cc6dd.

Best,
Martin

AndrewSB:
understood, yeah those changes look like they'll work great for me :)

PR #141, opened 2022-07-19
MessageView: make the custom padding customizable one side at a time

AndrewSB:
continuing from https://github.com/GetStream/stream-chat-swiftui/pull/140

turns out the fix you've put in wasn't customizable enough, in the case i'm looking at right now, i only have to customize the `topPadding` of a `MessageTextView` to be 0 when there's a quoted message above it.


martinmitrevski (approved):
Thanks @AndrewSB, this makes sense 👍

PR #142, opened 2022-07-19
MessageDisplayOptions: make showing avatars configurable between DMs or groups

AndrewSB:
we wanted to show avatars in group chats, but not DMs


martinmitrevski:
This is kind of a breaking change, since `showAvatars` is public and used by some customers. We will pick it up internally for further analysis.

PR #143, opened 2022-07-20
stop building channel selection info according channel name which is not uniq

### LINKED ISSUE #122: Getting booted out of channel after editing channel name

saidReclip:
this fixes #122

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
@@ -81,7 +81,7 @@ public struct ChannelSelectionInfo: Identifiable {
         if let message = message {
             id = "\(channel.cid.id)-\(message.id)"
         } else {
-            id = channel.name ?? channel.cid.id
+            id = channel.cid.id
         }
     }
 }


martinmitrevski (approved):
Thanks @saidReclip, this is a great find! 

@adolfogarza FYI

PR #153, opened 2022-07-27
Add a shadow for a play icon

saidReclip:
Hi!
The play icon is not visible enough on white backgrounds. 
Added a shadow.


martinmitrevski (approved):
thanks @saidReclip 👍

PR #158, opened 2022-08-02
Create new factory view: MessageTextContentView

AndrewSB:
we needed this to set the line height of the MessageTextView. I thought being allowed to change this as a factory method was easier than adding more parameters, also we wanted to take advantage of your quoted message cache instead of checking fresh each time

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
@@ -145,7 +145,8 @@ public struct MessageTextView<Factory: ViewFactory>: View {
             alignment: message.alignmentInBubble,
             spacing: 0
         ) {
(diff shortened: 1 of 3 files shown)


martinmitrevski:
Hey @AndrewSB,

Thanks for the PR. In this case, I don't think adding additional extension point is necessary, because of the following:
- might be confusing for customers, since we already expose the message text view. The difference is very small and might lead to confusion.
- customizing the message text view is fairly simple - there are few lines of code and everything is public.
- regarding the `quotedMessage`, you can safely use it directly from the message. Our performance analysis have shown that only the message.author improves performance (since it's repeatable). Therefore, we might even remove the cached quoted message at some point. 

Looking forward to your response.

AndrewSB:
got it, if there's no real performance gain to using the quotedMessage from your cache, we'll just override the MessageTextView itself

PR #168, opened 2022-08-05
fix header flickering

saidReclip:
Bug: Header flickers

STR's: 
1. open a channel
2. receive many messages from channel participant
3. sometimes you will see header redrawing 

https://user-images.githubusercontent.com/102264702/183126227-13b6691a-df0c-46d7-b6d2-5a53aa13ee12.mp4

this change should fix header flickering


martinmitrevski:
Hey @saidReclip,

Thanks for the PR. The fix you've provided actually breaks typing indicators in the nav bar. 

However, we've applied a different fix that addresses the root issue. This is now available on our `main` branch.

Best,
Martin

PR #206, opened 2022-09-19
Said/channel attachments minor fixes

saidReclip:
Hi there, 
We need some of this struct be public for our customization. 
Is it okay to merge this things? 
I also added shadows to VideoIndicatorView and VideoDurationIndicatorView to make them more visible on white background.


martinmitrevski (approved):
Thanks @saidReclip, we can include these changes.

PR #207, opened 2022-09-19
Said/becomes first responder on open

saidReclip:
### 🎯 Goal
Make an opportunity to open a channel thread with becomesFirstResponder true or false when it is needed independently of default value from config. 

### 🛠 Implementation

Add an optional becomesFirstResponderOnOped to navigation cycle


saidReclip:
Hi @martinmitrevski. Thank you for your response. 
What if we will make this argument nil by default? It will not break Factory for customers in this case. 
Is it possible to merge then?

martinmitrevski:
Hey @saidReclip, 
Default argument values are not allowed for protocols in Swift. Our `ViewFactory` is a protocol, and the customers provide their own implementations. Even if we add default values in our `DefaultViewFactory`, we will still break our customers' implementations.

saidReclip:
Yeah. Forget about it. It will break then.

Issue #219, opened 2022-10-17
Remove / Make-Optional the root NavigationView


Sam-Spencer:
## What are you trying to achieve?
`ChatChannelListView` has its `body` content wrapped inside a `NavigationView`. Without writing a completely custom `ChatChannelListView`, it is currently impossible to push this view onto an existing navigation stack. It would be nice if this `NavigationView` wrapping was optional.

## If possible, how can you achieve this currently?
It's impossible to push the `ChatChannelListView` onto any existing navigation stack in SwiftUI. If you do, SwiftUI presents all subsequently pushed content with **two(!)** navigation bars. Yuck! 🤮 

The easiest "workaround" is to present the `ChatChannelListView` as a `sheet`. But that's unfortunate because chat experiences don't usually lend themselves to modal-style presentations.

Another potential workaround is to completely re-implement the entire `ChatChannelListView` yourself. Also, not a great option if you're trying to plug it in and _go_.

## What would be the better way?
I understand the importance of having the `NavigationView` as the root view for ease-of-use, but it may be invaluable to provide an additional hook for `ViewFactory` subclasses. Separating out the enclosed content from the `NavigationView` and allowing a `ViewFactory` to specify the root behavior would solve this issue while maintaining ease-of-use.

So instead of this (current implementation):

```swift
public var body: some View {
    NavigationView {
        channelListViewContent() // Just using this pseudo-function to stand in for the actual contents
    }
}
```

You might have something like this:

```swift
public var body: some View {
    if viewModel.embedInNavView == true {
        NavigationView {
            channelListViewContent()
        }
    } else {
        channelListViewContent()
    }
}
```

Possible solution in PR #220 

## GetStream Environment
**GetStream Chat version:** 4.22.0
**GetStream Chat frameworks:** StreamChat, StreamChatSwiftUI
**iOS version:** 16.0
**Swift version:** 5.7
**Xcode version:** 14.0.1
**Device:** iPhone 14 Pro

## Additional context
N/A

martinmitrevski:
Hey @Sam-Spencer,

Thanks for reporting. As discussed, this will be handled with https://github.com/GetStream/stream-chat-swiftui/pull/220.

Best,
Martin

PR #220, opened 2022-10-17
Allow opting-out of NavigationView embed for ChatChannelListView

Sam-Spencer:
### 🔗 Issue Link
GitHub Issue #219

### 🎯 Goal
Allow for more flexible presentation of the `ChatChannelListView`. In its current form, the messages list view can only be displayed as a root view (e.g. sidebar or tab item) or via a modal-presentation (e.g. `sheet` or `fullScreenCover`). Currently, if you attempt to push the `ChatChannelListView` onto an existing navigation stack you'll wind up with two navigation bars layered vertically on the screen. This isn't right.

The goal of this PR is to preserve the usability and ease-of-use for existing and recommended implementations, while providing flexibility to present in the way that makes the most sense for an application.

### 🛠 Implementation
To achieve this, I've added a flag to the `MessageListConfig` object: `embedInNavigationView`. The new value defaults to `true`, ensuring no breaking changes to existing behavior. However, when set to `false` the `ChatChannelListView` will be presented without an enclosing `NavigationView`. In that case, any presenting view should then take the responsibility of wrapping it in an appropriate navigation mechanism.

### 🧪 Testing
Testing is easy! Run as is to ensure the project still behaves as expected. You can run the Demo app to verify as well as UI and Unit tests.

### ☑️ Checklist

- [x] New code is covered by unit tests -- *Nothing to change*
- [x] Affected documentation updated (docusaurus, tutorial, CMS (task created)) -- *I think I did this properly*


Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListConfig.swift
```diff
@@ -54,6 +56,7 @@ public struct MessageListConfig {
     public let maxTimeIntervalBetweenMessagesInGroup: TimeInterval
     public let cacheSizeOnChatDismiss: Int
     public let iPadSplitViewEnabled: Bool
+    public let embedInNavigationView: Bool
```
martinmitrevski, line 59: The `MessageListConfig` is mostly about configs related to the message list. We should move the `embedInNavigationView` as a parameter in the initializer of the `ChatChannelListView`, with the default value of true.

docusaurus/docs/iOS/swiftui/chat-channel-components/message-list.md
```diff
@@ -302,6 +303,12 @@ let utils = Utils(messageListConfig: messageListConfig)
 streamChat = StreamChat(chatClient: chatClient, utils: utils)
 ```
 
+### embedInNavigationView
```
martinmitrevski, line 306: This would need to be moved to the docs about the channel list view.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
```diff
@@ -45,72 +45,86 @@ public struct ChatChannelListView<Factory: ViewFactory>: View {
...
+    }
+    
+    @ViewBuilder
+    private func container() -> some View {
+        if utils.messageListConfig.embedInNavigationView == true {
+            NavigationView {
+                content()
+            }
+        } else {
+            content()
```
martinmitrevski, line 80: Would be nice to have a snapshot test for this case.

martinmitrevski (commented):
Hey @Sam-Spencer, 

Thanks for taking the time to create this PR, we really appreciate your efforts.
This is a nice improvement, that we will include in the SDK. There are few smaller things that I've added as comments, that we should fix before merging. 

Let us know if you have the time to address those, or we should pick it up from here?

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListConfig.swift
Sam-Spencer: Moved this to the initializer of `ChatChannelListView` (as requested)!

docusaurus/docs/iOS/swiftui/chat-channel-components/message-list.md
Sam-Spencer: Updated documentation  to reflect recent changes

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
Sam-Spencer: Added a snapshot test case. Not 100% sure if I did this right, but the snapshot looks correct and the test case passes.

martinmitrevski (approved):
thanks @Sam-Spencer, great work 👏

PR #226, opened 2022-11-02
add lineSpacing to MessageTextView

saidReclip:
Hi! 
We need to this to customize MessageTextView.

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
@@ -118,6 +118,7 @@ public struct MessageTextView<Factory: ViewFactory>: View {
     private let trailingPadding: CGFloat
     private let topPadding: CGFloat
     private let bottomPadding: CGFloat
+    private let lineSpacing: CGFloat
     @Binding var scrolledId: String?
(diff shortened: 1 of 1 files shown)


Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
```diff
@@ -128,6 +129,7 @@ public struct MessageTextView<Factory: ViewFactory>: View {
         trailingPadding: CGFloat = 16,
         topPadding: CGFloat = 8,
         bottomPadding: CGFloat = 8,
+        lineSpacing: CGFloat = 1,
```
martinmitrevski, line 132: I couldn't find any official docs about the default value of `lineSpacing`. Not sure if it's 1, and if that applies for every dynamic font size. Therefore, there's a risk that this might be a breaking change.

martinmitrevski (commented):
I've added one comment about the default value of `lineSpacing`. This might be a breaking change for us. If you have some more info how you got this value (and if it impacts dynamic font sizes), please share it.

martinmitrevski:
As discussed, we can't merge this, since the lineSpacing's default value is not documented, and might trigger some unexpected side-effects.

PR #231, opened 2022-11-09
Bump nuke

AndrewSB:


martinmitrevski:
Hey @AndrewSB,

Thanks for creating the PR. We bumped it internally, since there were some other places where we needed to update this.

It's already available on `main` and it will be included in the next release.

PR #235, opened 2022-11-24
add offset Y

saidReclip:
### 🎯 Goal

We need to observe offset Y of ChannelList

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
@@ -15,6 +15,7 @@ public struct ChannelList<Factory: ViewFactory>: View {
     var channels: LazyCachedMapCollection<ChatChannel>
     @Binding var selectedChannel: ChannelSelectionInfo?
     @Binding var swipedChannelId: String?
+    @Binding var offsetY: CGFloat?
     private var scrollable: Bool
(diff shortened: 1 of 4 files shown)


saidReclip:
Hello @nuno-vieira! 
Could we get a review on this?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
```diff
@@ -56,13 +58,24 @@ public struct ChannelList<Factory: ViewFactory>: View {
...
         _swipedChannelId = swipedChannelId
+        _offsetY = offsetY
     }
     
     public var body: some View {
         Group {
             if scrollable {
                 ScrollView {
-                    channelsVStack
+                    ZStack {
```
martinmitrevski, line 68: Thanks for the PR. Not sure if this should be enabled by default for all users. We should probably have a config with an opt-in. Will think about this and do a follow-up.

nuno-vieira:
> Hello @nuno-vieira! Could we get a review on this?

@saidReclip, @martinmitrevski will do it when he has availability

martinmitrevski:
Hey @saidReclip,

Thanks again for creating the PR. Unfortunately, this introduces noticeable performance degradation in the scrolling experience (especially with many channels), therefore we can't merge it in. 

Instead, I've made the `ChannelsLazyVStack` public, so you can do this in your app (in case you don't have these performance issues).

Best,
Martin

PR #285, opened 2023-03-03
Expose SearchBar and add customizable placeholder

grandsir:
this will make `makeChannelListTopView` more customizable, because, in our app, we needed to add a plus button near the searchbar, but we had to recreate the component just because it is internal.

### 🎯 Goal

Making SearchBar accessible

### 🛠 Implementation

-

### 🧪 Testing

-

### 🎨 Changes

-

### ☑️ Checklist

- ☑️ I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)

### DIFF
--- a/Sources/StreamChatSwiftUI/CommonViews/SearchBar.swift
+++ b/Sources/StreamChatSwiftUI/CommonViews/SearchBar.swift
@@ -5,7 +5,8 @@
 import SwiftUI
 
 /// Search bar used in the message search.
-struct SearchBar: View, KeyboardReadable {
+public struct SearchBar: View, KeyboardReadable {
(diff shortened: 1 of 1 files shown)


martinmitrevski:
Hey @grandsir,

Thanks for the PR, much appreciated. However, we are keeping the `SearchBar` private on purpose - it's a custom implementation, since SwiftUI doesn't offer a native component. 
We would like to keep the possibility to replace this implementation if/when SwiftUI offers something out of the box, which would not be possible if it's a public component.
Therefore, the recommended way for now is to re-implement it by yourself.
Best, Martin

Issue #301, opened 2023-04-06
Support Nuke 12


paulz:
## What are you trying to achieve?
App depends on Nuke >12.0.0

## If possible, how can you achieve this currently?
Downgrade to Nuke 9

## What would be the better way?
v 4.30.0 allows Nuke 12

## GetStream Environment
**GetStream Chat version:** 4.3.0
**GetStream Chat frameworks:** StreamChat, StreamChatSwiftUI
**iOS version:** 16.0
**Swift version:** 5
**Xcode version:** 14.2
**Device:** iPhone

## Additional context

Xcode fails to resolve dependencies:

> Showing All Errors Only
> Failed to resolve dependencies Dependencies could not be resolved because root depends on 'nuke' 12.0.0..<13.0.0 and root depends on 'stream-chat-swiftui' 4.0.0..<5.0.0.
> 'stream-chat-swiftui' >= 4.0.0 practically depends on 'nuke' {10.0.0..<11.0.0, 11.3.0} because 'stream-chat-swiftui' 4.29.0 depends on 'nuke' 11.3.1 and 'stream-chat-swiftui' 4.30.0 depends on 'nuke' 11.3.1.
> 'stream-chat-swiftui' {4.0.0..<4.29.0, 4.29.1..<4.30.0, 4.30.1..<5.0.0} practically depends on 'nuke' {10.0.0..<11.0.0, 11.3.0} because 'stream-chat-swiftui' 4.27.0 depends on 'nuke' 11.3.1 and 'stream-chat-swiftui' 4.28.0 depends on 'nuke' 11.3.1.
> 'stream-chat-swiftui' {4.0.0..<4.27.0, 4.27.1..<4.28.0, 4.28.1..<4.29.0, 4.29.1..<4.30.0, 4.30.1..<5.0.0} practically depends on 'nuke' {10.0.0..<11.0.0, 11.3.0} because 'stream-chat-swiftui' 4.25.0 depends on 'nuke' 11.3.1 and 'stream-chat-swiftui' 4.26.0 depends on 'nuke' 11.3.1.
[...]

PR #341, opened 2023-07-12
Package.swift: upgrade to Nuke 12

AndrewSB:


martinmitrevski:
Thanks for the PR. As discussed here: https://github.com/GetStream/stream-chat-swiftui/issues/301, we can't upgrade the Nuke version at the moment.

PR #369, opened 2023-09-27
Update .spi.yml

finestructure:
We're about to switch documentation generation to 5.9 by default. Our regression test suite is flagging a doc gen failure for stream-chat-swiftui when running with 5.9:

```
SwiftDocC/DocumentationContext.swift:1206: Fatal error: Symbol with identifier 's:10StreamChat19MessageReactionTypeV0aB7SwiftUIE2idSSvp' has no reference. A symbol will always have at least one reference.
Command CompileDocumentation failed with a nonzero exit code
```

This change pins the documentation generation to 5.8 for now.

Steps to reproduce:

Either just via Product → Build Documentation in Xcode 15 or as follows (which is essentially what we do to generate docs):

```
git clone https://github.com/GetStream/stream-chat-swiftui/
cd stream-chat-swiftui/
git checkout 4.37.0
env DEVELOPER_DIR=/Applications/Xcode-15.0.0.app xcrun \
            xcodebuild \
            -IDEClonedSourcePackagesDirPathOverride="$PWD/.dependencies" \
            -derivedDataPath "$PWD/.derivedData" \
            docbuild \
            OTHER_DOCC_FLAGS=--emit-digest \
            -scheme StreamChatSwiftUI \
            -destination 'generic/platform=ios'
```

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -4,3 +4,4 @@ builder:
   - platform: ios
     documentation_targets: [StreamChatSwiftUI]
     scheme: StreamChatSwiftUI
+    swift_version: 5.8


finestructure:
cc @franklinsch @ethan-kusters because this working on 5.8 but failing on 5.9 might be a regression in DocC itself

martinmitrevski (approved):
thanks @finestructure for taking the time to fix this 🙏

QuietMisdreavus:
@finestructure I took a look, and it seems like this passes in Xcode 15 with "Include Documentation for Symbols with Swift Extensions" turned off (`DOCC_EXTRACT_EXTENSION_SYMBOLS=NO`), so you could try that if you need to use anything from Swift 5.9. I'm going to investigate an upstream fix for this.

finestructure:
Thanks a lot for taking a look, @QuietMisdreavus ! We don't support that setting as a user-configurable option at the moment. Are you suggesting it's a setting we should apply by default across all doc builds? Or just something specific to this package? (Maybe something to discuss at the next DocWG meeting!)

QuietMisdreavus:
I've seen it in a few places so far, so it may need to be set universally for now. 😞 We can definitely bring it up at the WG meeting to bring in more opinions, though.

finestructure:
I see, that's good to know - thanks! I've added an issue over in SPI-Server to track this on our end.

PR #430, opened 2024-02-07
ViewFactory: add variant of makeMessageListContainerModifier with channel

AndrewSB:
i've just added a new view factory function with a channel param. shouldn't be a breaking change

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListView.swift
@@ -304,6 +304,7 @@ public struct MessageListView<Factory: ViewFactory>: View, KeyboardReadable {
                 ) : nil
         )
         .modifier(factory.makeMessageListContainerModifier())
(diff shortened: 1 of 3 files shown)


martinmitrevski:
Hey @AndrewSB, thanks for raising the PR. It's not a breaking change, but it will be confusing for our customers to have 2 versions of the same method and long term it won't be scalable for us to maintain that. Therefore, we will close this PR, but will keep it in mind for a major version release.

AndrewSB:
got it, i'm going to have to maintain a fork until this is included upstream, so i'd really appreciate this making it into your version

martinmitrevski:
one idea to overcome this: keep the currently selected channel as a var in your factory and then pass it in the modifier method we already have.

AndrewSB:
could do, but i'm worried about some sort of edge case where we don't clear that channel/we deeplink into another channel and don't update the var in time

AndrewSB:
@martinmitrevski what if we added an `unsafe_` prefix to this function? 

it would technically enter your API compatibility, but with the unsafe prefix, i think it would be totally fair for you to break the API once there's a major release?

PR #475, opened 2024-04-16
Do not access local state when updating the state in sink()

laevandus:
### 🔗 Issue Link

Related: [#728](https://github.com/GetStream/ios-issues-tracking/issues/728)

### 🎯 Goal

`sink()` is triggered when the observed property calls willSet (@Published property wrapper's default behaviour), therefore we should use the received value for state updates.

### 🛠 Implementation

Other alternatives: 
- use `receive(on)` for delaying the update
- assign the observed value to a local property first (used in some cases)

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
@@ -83,7 +83,8 @@ import SwiftUI
         didSet {
             // When reactions are shown, the navigation bar is hidden.
             // Check the header type and trigger an update.
-            checkHeaderType()
+            guard let channel else { return }
+            checkHeaderType(for: channel)
         }
     }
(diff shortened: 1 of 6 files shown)


Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -398,12 +401,12 @@ import SwiftUI
     
     func dataSource(
         channelDataSource: ChannelDataSource,
-        didUpdateChannel channel: EntityChange<ChatChannel>
+        didUpdateChannel channel: ChatChannel
```
laevandus, line 404: I ended up changing the code to use the received value and not accessing `self.channel` which would still point at the previous value.

martinmitrevski (approved):
LGTM ✅

PR #476, opened 2024-04-16
[New state] Set the initial state for read, typing and online indicators

laevandus:
### 🔗 Issue Link
Related: [#728](https://github.com/GetStream/ios-issues-tracking/issues/728)

### 🎯 Goal

Fix "If you open the channel when participant was typing, you won’t see typing indicator"

### 🛠 Implementation

We are not setting the initial state correctly.

```swift
  channelDataSource = ChatChannelDataSource(chat: chat)
}
channelDataSource.delegate = self
```

```swift
@MainActor init(chat: Chat) {
        self.chat = chat
        subscribeForMessageUpdates()
        subscribeForChannelUpdates()
    }

 @MainActor private func subscribeForChannelUpdates() {
        self.chat.state.$channel.sink { [weak self] (channel: ChatChannel?) in
            guard let self else { return }
            if let channel {
                delegate?.dataSource(
                    channelDataSource: self,
                    didUpdateChannel: channel
                )
            }
        }
        .store(in: &cancellables)
    }
```
Here we create the data source, start observing it, but since delegate is nil, the initial delegate callback does not happen and we miss the initial state.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
@@ -185,9 +185,6 @@ import SwiftUI
         }
                 
         channelName = channel?.name ?? ""
-        if let channel {
-            checkHeaderType(for: channel)
-        }
         checkUnreadCount()
     }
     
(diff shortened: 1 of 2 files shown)


Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -185,9 +185,6 @@ import SwiftUI
         }
                 
         channelName = channel?.name ?? ""
-        if let channel {
```
laevandus, line 188: Triggered when channel changes, so no need to call it twice

PR #495, opened 2024-05-27
Implement polls feature

martinmitrevski:
### 🔗 Issue Link
_Jira or Github issue link, if applicable._

### 🎯 Goal

### 🛠 Implementation

_Provide a description of the implementation._

### 🧪 Testing

_Describe the steps how this change can be tested (or why it can't be tested)._

### 🎨 Changes

### ☑️ Checklist


Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -177,6 +177,20 @@ open class ChatChannelViewModel: ObservableObject, MessagesDataSource {
             object: nil
         )
         
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(onViewAppear),
+            name: NSNotification.Name(.sheetHiddenNotification),
+            object: nil
+        )
```
laevandus, line 185: What was the issue this code is fixing? Currently it is just sent by some sheets from poll attachment view but there are other sheets as well which do not send any notifications.

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelDataSource.swift
```diff
@@ -161,7 +161,11 @@ class MessageThreadDataSource: ChannelDataSource, ChatMessageControllerDelegate
     weak var delegate: MessagesDataSource?
     
     var messages: LazyCachedMapCollection<ChatMessage> {
-        messageController.replies
+        var replies = messageController.replies
+        if let message = messageController.message, replies.last != message {
+            replies.append(message)
+        }
+        return replies
```
laevandus, line 168: This fixed an issue where showing a sheet and dismissing it, triggered channel view model to read data source's messages which dropped the parent message. 
e.g. use thread reply for polls, in the thread, open a sheet > poll disappears

Sources/StreamChatSwiftUI/ChatChannel/MessageList/Polls/PollAttachmentView.swift
```diff
@@ -0,0 +1,271 @@
...
+                
+                HStack {
+                    Text(subtitleText)
+                        .font(fonts.caption1)
+                        .foregroundColor(Color(colors.textLowEmphasis))
+                    Spacer()
+                }
+            }
+            
+            ForEach(options.prefix(10)) { option in
```
laevandus, line 54: 🤔  10 could be a constant since it is used in multiple places and has potential of being configurable value

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
martinmitrevski: fixes the sheet disappearing when adding reactions. Not sure if it's possible to fix it in another way, given the way SwiftUI works. The reactions would change the id of the message, thus changing its identity. With the fix, when there's a sheet open, we delay the update of the message.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/Polls/PollAttachmentView.swift
martinmitrevski: sure, feel free to add it

Sources/StreamChatSwiftUI/ChatChannel/MessageList/Polls/PollsHelpers.swift
```diff
@@ -0,0 +1,62 @@
...
+            .font(fonts.subheadline)
+            .foregroundColor(Color(colors.textLowEmphasis))
+    }
+}
+
+class PollsDateFormatter {
+    
+    let dateFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.setLocalizedDateFormatFromTemplate("MMM dd HH:mm")
```
laevandus, line 49: @martinmitrevski I changed this to localized date format. Otherwise we do not respect the month and day order in different languages.

old

new

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
laevandus: I see the problem now. Maybe we should rename the notification to be message specific. `.messageCellSheetHiddenNotification` or smt which tells that this is related to the message list only.

laevandus: I will document this.

laevandus (approved):

PR #508, opened 2024-06-07
Bump min supported swift version

testableapple:


.github/workflows/smoke-checks.yml
```diff
@@ -223,7 +223,7 @@ jobs:
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GITHUB_PR_NUM: ${{ github.event.number }}
-      XCODE_VERSION: "14.0.1"
+      XCODE_VERSION: "15.0.1"
```
laevandus, line 226: 15.4?

testableapple: less supported is 15.0.1 and this is the one we use in `build-xcode15` job

laevandus (approved):
Let's verify that Package.swift does not generate any warnings since sometimes Apple deprecates stuff in different versions.

testableapple:
> Let's verify that Package.swift does not generate any warnings since sometimes Apple deprecates stuff in different versions.

Looks fine to me

PR #515, opened 2024-06-12
Reload messages when author name or profile picture changes

laevandus:
### 🔗 Issue Link

Resolves [ios-issues-tracking/issues/861](https://github.com/GetStream/ios-issues-tracking/issues/861)

### 🎯 Goal

Reload messages when changing current user's name or profile picture.

### 🛠 Implementation

* Add author information to the ChatMessage.messageId which is used for reloading views through explicit id
* Reset cache when message list changes since any property of the message can change

### 🧪 Testing

We should play around with channel view a bit to see if things work normally. The reason is that we flush cached values more often.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
@@ -391,6 +391,8 @@ open class ChatChannelViewModel: ObservableObject, MessagesDataSource {
         if !isActive {
             return
         }
+        // Message data can change with every callback
+        utils.messageCachingUtils.clearCache()
         
         let animationState = shouldAnimate(changes: changes)
         if animationState == .animated {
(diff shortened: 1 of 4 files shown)


Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -391,6 +391,8 @@ open class ChatChannelViewModel: ObservableObject, MessagesDataSource {
         if !isActive {
             return
         }
+        // Message data can change with every callback
+        utils.messageCachingUtils.clearCache()
```
martinmitrevski, line 395: Not sure about this one. Another option could be to listen to specific events via an events controller?

laevandus: Thinking about closing this PR since the ticket about improving message list updates reverses these changes here. I think it is going to be better if I tackle this bug after those changes.

PR #525, opened 2024-06-25
Grouping image and video attachments in the same message

martinmitrevski:
### 🔗 Issue Link
Resolves https://stream-io.atlassian.net/browse/PBE-4343?atlOrigin=eyJpIjoiNWE3NzBlZThiZDUwNGNiM2IzNTRlODBkYmI5ZTFiZDQiLCJwIjoiaiJ9.

### 🎯 Goal

Images and videos should be part of the same message.

### 🛠 Implementation

Refactored the existing image attachment view to support videos. Kept the existing ones for backward compatibility.

### 🧪 Testing

Try different combinations of images and videos (and single ones too).

### 🎨 Changes

### ☑️ Checklist


laevandus:
I tried the branch and works well. 
What I observed was that video attachments are rendered as first items always in-spite of the order how I picked them (pick images first, then videos).

laevandus (approved):

martinmitrevski:
> I tried the branch and works well. What I observed was that video attachments are rendered as first items always in-spite of the order how I picked them (pick images first, then videos).

Yes, that's fine for now. Otherwise we would need to fetch all type erased attachments and cast them, so it might slow things down. We can revisit if there are complaints.

PR #534, opened 2024-07-02
Jumping to an unread message did not keep scroll location

laevandus:
### 🔗 Issue Link
Resolves [PBE-5003](https://stream-io.atlassian.net/browse/PBE-5003)

### 🎯 Goal

Fix an issue where scroll location was not kept after jumping to a message

### 🛠 Implementation

- Set the scroll location to the unread message when the message list has changed (unread message lookup only works on the loaded message list)

Note: In general, the ChatChannelViewModel would benefit from the state layer API a lot since then we would not have cases where we call load more messages, wait for completion handler AND then also need to wait when the delegate callback comes (delay comes from database observer).

### 🧪 Testing

Chewbacca has some test channels useful here.

Verification checks
1. Is scroll to last button shown or hidden correctly?
2. Is the scrolled message visible after jumping to it?
3. Unread messages should be marked as read when reaching the most recent message after jumping to the first unread message.

Case 1: Jumping to unread message
1. Scroll up in a channel and press and hold on a message
4. Mark the message as unread
5. Go to channel list
6. Go back to the channel
7. Tap on the x unread messages button > we should jump to the first unread message

Case 2: Jumping to some other message
1. Scroll to an old message, quote it
2. Open the channel again and tap on the quoted message > we jump to that message

### 🎨 Changes

I marked the message "A" as unread.

| Before | After |
| ------ | ----- |
| img   |  img  |

### ☑️ Checklist

[PBE-5003]: https://stream-io.atlassian.net/browse/PBE-5003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
@@ -44,6 +44,7 @@ open class ChatChannelViewModel: ObservableObject, MessagesDataSource {
     
     private var loadingPreviousMessages: Bool = false
     private var loadingMessagesAround: Bool = false
+    private var scrollsToUnreadAfterJumpToMessage = false
     private var lastMessageRead: String?
     private var disableDateIndicator = false
     private var channelName = ""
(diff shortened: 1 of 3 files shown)


Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListView.swift
```diff
@@ -230,8 +230,8 @@ public struct MessageListView<Factory: ViewFactory>: View, KeyboardReadable {
                             } else if diff < 0 && scrollDirection == .down {
                                 scrollDirection = .up
                             }
+                            utils.messageCachingUtils.scrollOffset = offsetValue
```
laevandus, line 233: Here we had an issue where if you scroll extra slowly, the scroll direction argument in the view model is incorrect.

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -472,31 +472,29 @@ open class ChatChannelViewModel: ObservableObject, MessagesDataSource {
...
-                limit: utils.messageListConfig.pageSize,
-                completion: { [weak self] _ in
-                    guard let self = self else { return }
+        loadingPreviousMessages = true
+        channelDataSource.loadPreviousMessages(
+            before: nil,
+            limit: utils.messageListConfig.pageSize,
+            completion: { [weak self] _ in
+                guard let self = self else { return }
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
laevandus, line 486: We could delete all of the 0.5 second delay logic if we switch to state layer's Chat but that is another topic.

martinmitrevski (commented):
there's still one issue here - when you mark unread an older message (e.g. unread count > 30), and you come back, sometimes it marks the chat as read.

martinmitrevski (approved):
Works well now ✅

PR #535, opened 2024-07-03
Skip pod lint when targeting a branch for stream-chat-swift

laevandus:
### 🎯 Goal

Keep pod linting green on CI when targeting a branch during development.

### 🛠 Implementation

Often we use StreamChat's develop branch in development. This triggers CI errors when linting the pod file. Skip linting if we target a branch.

Example: [CI fails](https://github.com/GetStream/stream-chat-swiftui/actions/runs/9773393802/job/26979475151?pr=534)

### ☑️ Checklist

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -15,7 +15,7 @@ derived_data_path = 'derived_data'
 source_packages_path = 'spm_cache'
 buildcache_xcargs = 'CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++'
 is_localhost = !is_ci
-project_package_resolved = './../StreamChatSwiftUI.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved'
+project_package_resolved = "#{xcode_project}/project.xcworkspace/xcshareddata/swiftpm/Package.resolved"
 @force_check = false
 
 before_all do |lane|
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -127,7 +127,7 @@ end
 
 lane :pod_lint do
   dependencies = Dir.chdir('..') { JSON.parse(File.read(project_package_resolved)) }
-  is_branch = dependencies["pins"].find { |dependency| dependency['identity'] == 'stream-chat-swift' && dependency['state']['branch'] }
+  is_branch = dependencies['pins'].find { |dependency| dependency['identity'] == 'stream-chat-swift' && dependency['state']['branch'] }
   if is_branch.nil?
     pod_lib_lint(podspec: 'StreamChatSwiftUI.podspec', allow_warnings: true)
   else
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -126,7 +126,7 @@ private_lane :appstore_api_key do
 end
 
 lane :pod_lint do
-  dependencies = JSON.parse(File.read(project_package_resolved))
+  dependencies = Dir.chdir('..') { JSON.parse(File.read(project_package_resolved)) }
   is_branch = dependencies["pins"].find { |dependency| dependency['identity'] == 'stream-chat-swift' && dependency['state']['branch'] }
   if is_branch.nil?
     pod_lib_lint(podspec: 'StreamChatSwiftUI.podspec', allow_warnings: true)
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -15,6 +15,7 @@ derived_data_path = 'derived_data'
 source_packages_path = 'spm_cache'
 buildcache_xcargs = 'CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++'
 is_localhost = !is_ci
+project_package_resolved = './../StreamChatSwiftUI.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved'
```
testableapple, line 18: `./../` might be confusing/not available in some cases. I think better to use the root folder path and check it out where required via `Dir.chdir('..')` as we usually do.

```suggestion
project_package_resolved = "#{xcode_project}/project.xcworkspace/xcshareddata/swiftpm/Package.resolved"
```

fastlane/Fastfile
```diff
@@ -125,7 +126,13 @@ private_lane :appstore_api_key do
 end
 
 lane :pod_lint do
-  pod_lib_lint(podspec: 'StreamChatSwiftUI.podspec', allow_warnings: true)
+  dependencies = JSON.parse(File.read(project_package_resolved))
```
testableapple, line 129:
```suggestion
  dependencies = Dir.chdir('..') { JSON.parse(File.read(project_package_resolved)) }
```

fastlane/Fastfile
```diff
@@ -125,7 +126,13 @@ private_lane :appstore_api_key do
...
+  is_branch = dependencies["pins"].find { |dependency| dependency['identity'] == 'stream-chat-swift' && dependency['state']['branch'] }
```
testableapple, line 130:
```suggestion
  is_branch = dependencies['pins'].find { |dependency| dependency['identity'] == 'stream-chat-swift' && dependency['state']['branch'] }
```

fastlane/Fastfile
```diff
@@ -125,7 +126,13 @@ private_lane :appstore_api_key do
...
+  if is_branch.nil?
+    pod_lib_lint(podspec: 'StreamChatSwiftUI.podspec', allow_warnings: true)
+  else
+    UI.important('Pod linting skipped: stream-chat-swift targets a development branch')
+  end
```
testableapple, line 135: I'm curious, does the current solution mean that we have to build the package first to generate `Package.resolved` to check if `pod lint` is required? As far as I know, `Package.resolved` is in `.gitignore` and is not pushed to `origin/`.

If so, maybe we could just parse `StreamChatSwiftUI.xcodeproj/project.pbxproj` via [xcodeproj](https://github.com/CocoaPods/Xcodeproj) and read the dependency rule straight away? 🤔

testableapple (changes requested):

fastlane/Fastfile
laevandus: Did not figure it out so I added resolve packages step instead.

fastlane/Fastfile
```diff
@@ -125,7 +126,14 @@ private_lane :appstore_api_key do
...
 lane :pod_lint do
-  pod_lib_lint(podspec: 'StreamChatSwiftUI.podspec', allow_warnings: true)
+  Dir.chdir('..') { sh("xcodebuild -resolvePackageDependencies -clonedSourcePackagesDirPath #{source_packages_path}") }
+  dependencies = Dir.chdir('..') { JSON.parse(File.read(project_package_resolved)) }
+  is_branch = dependencies['pins'].find { |dependency| dependency['identity'] == 'stream-chat-swift' && dependency['state']['branch'] }
+  if is_branch.nil?
+    pod_lib_lint(podspec: 'StreamChatSwiftUI.podspec', allow_warnings: true)
+  else
+    UI.important('Pod linting skipped: stream-chat-swift targets a development branch')
+  end
```
testableapple, line 136: Looks good, but would be great to avoid double `Dir.chdir`ing. What do you thing about something like this?
```suggestion
  lint_required = true
  Dir.chdir('..') do
    sh("xcodebuild -resolvePackageDependencies -clonedSourcePackagesDirPath #{source_packages_path}")
    dependencies = JSON.parse(File.read(project_package_resolved))
    lint_required = dependencies['pins'].none? { |dependency| dependency['identity'] == 'stream-chat-swift' && dependency['state']['branch'] }
  end

  if lint_required
    pod_lib_lint(podspec: 'StreamChatSwiftUI.podspec', allow_warnings: true)
  else
    UI.important('Pod linting skipped: stream-chat-swift targets a development branch')
  end
```

testableapple (approved):
You're a Ruby master now ☀️

PR #607, opened 2024-09-20
Reduce view updates when computing frame for long press action

laevandus:
### 🎯 Goal

Reduce view updates when calculating a frame for the long press action (computing the frame led to a crash in rare cases)

### 🛠 Implementation

- onChange is called just after SwiftUI finishes view updates, therefore accessing the frame should not lead to in SwiftUI view graph calculations
- Update the boolean toggle just once, because it is not important if it stays in true or false (saving view update cycle)

### 🧪 Testing

Try opening a channel, sending messages, opening message actions etc

### 🎨 Changes

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
@@ -112,9 +112,7 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
                         GeometryReader { proxy in
                             Rectangle().fill(Color.clear)
                                 .onChange(of: computeFrame, perform: { _ in
-                                    DispatchQueue.main.async {
-                                        frame = proxy.frame(in: .global)
-                                    }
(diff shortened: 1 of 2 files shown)


Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -112,9 +112,7 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
                         GeometryReader { proxy in
                             Rectangle().fill(Color.clear)
                                 .onChange(of: computeFrame, perform: { _ in
-                                    DispatchQueue.main.async {
-                                        frame = proxy.frame(in: .global)
-                                    }
+                                    frame = proxy.frame(in: .global)
```
laevandus, line 115: Already on main thread and onChange is called just after SwiftUI finishes updating views, therefore should be safer to ask for the frame than delaying it. Not sure if it fixes the reported crash since I can't reproduce it, but worth the try.

PR #613, opened 2024-09-30
[CI] Improvements

testableapple:
- Extract `merge_main_to_develop` to an indie lane to be able to run it locally
- Trigger `release_publish.yml` workflow via `gh` instead of `push` event that did not work

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -1,10 +1,6 @@
 name: "Publish new release"
 
 on:
-  push:
-    branches:
-      - main
-
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


testableapple:
> Did you also update the release notion and the JIRA ticket?
Nothing changed yet. I still hope that granting Admin rights to bot will solve the issue. If not, I'll update the docs and remove this step from the pipeline.

martinmitrevski:
> > Did you also update the release notion and the JIRA ticket?
> 
> Nothing changed yet. I still hope that granting Admin rights to bot will solve the issue. If not, I'll update the docs and remove this step from the pipeline.

Adding admin rights to a bot could be risky - I would prefer if we stick to this step.

testableapple:
Ok, let's revert it back to `write` permissions and manual step.

PR #628, opened 2024-10-24
Add support for Channel Search in the Channel List

nuno-vieira:
### 🔗 Issue Link
Resolves https://stream-io.atlassian.net/browse/PBE-6220

### 🎯 Goal

Adds support for Channel Search in the Channel List.

**Usage:**

Provide a `searchType` of `.channels` when creating the `ChatChannelListView:`
```swift
ChatChannelListView(
   viewFactory: DemoAppFactory.shared,
   channelListController: channelListController,
   searchType: .channels
)
```

Docs PR: https://github.com/GetStream/stream-chat-swift/pull/3471

### 🛠 Implementation

A new `searchType` is introduced when creating a Channel List. The channel list view model will use this property to decide to perform a search by channels or by messages.

Another alternative approach was to introduce some kind of `ChatChannelListSearchResultsProvider`, which would abstract how the search implementation is done. However, the implementation was a bit more extensive and complex, so for now, this simple switch-case approach works fine and is easy to understand. In case we need a more scalable implementation in the future, we can consider this approach again. Here is the WIP implementation: https://github.com/GetStream/stream-chat-swiftui/commit/b4ac2df9b2b8e178cc816c8f0f93365a683790ec

### 🧪 Testing

Change the `DemoAppSwiftUIApp.channelListSearchType` to `.channels` in the code.

Since there is no Configuration Screen in the SwiftUI app to test this, it needs to be done using Xcode.

### ☑️ Checklist

- [x] Affected documentation updated (docusaurus, tutorial, CMS (task created)


nuno-vieira:
> Looks great! Just one question - also relates to the UIKit search. At the moment, this only works with channel names, wonder if we can extend it to channel member names? (But this is out of scope for this PR)

Yes this is also possible, is just a question of changing the autocomplete from `.name` to `.memberName` and then we can add an addition search type like `.channelMembers` 👍

spartan-tinhnguyen:
Hi, could we release this version soon? I need this feature so that my app can be released soon. Thanks!

nuno-vieira:
Hi @spartan-tinhnguyen,

The ETA is most likely next week.

spartan-tinhnguyen:
Hi @nuno-vieira , I’m sorry to ask this here, but could we also make this value public in the next version? 
`internal var messagePreviewFormatter = MessagePreviewFormatter()`
It is in Utils class.

spartan-tinhnguyen:
Hi @nuno-vieira 
I've applied the channel search, but it isn't working. Could you please help check it?
```
ChatChannelListView(
      viewFactory: CustomChatFactory.shared,
      embedInNavigationView: false,
      searchType: .channels
    )
```

Screenshot 2024-11-07 at 22 34 31
Screenshot 2024-11-07 at 22 35 07

nuno-vieira:
Hi @spartan-tinhnguyen,

If that is a DM channel, then it won't work, because the channel search, at the moment only searches for Channel Names. DM Channels do not have channel names

spartan-tinhnguyen:
@nuno-vieira  Do you have another way to customize the DM channel search?

nuno-vieira:
@spartan-tinhnguyen Can you create an issue here on the Repo? So that we can discuss it there?

PR #639, opened 2024-11-04
Fix Channel List stuck in Empty View State in rare conditions

nuno-vieira:
### 🔗 Issue Link
Resolves https://stream-io.atlassian.net/browse/PBE-6291

### 🎯 Goal

Fix the Channel List stuck in Empty View State in rare conditions.

### 🛠 Implementation

- Fix the Channel List stuck in Empty View State in rare conditions. When a customer selected a channel, before opening the channel, the empty view would not disappear.
- Fixes the loading view not showing in rare scenarios, like in the simulator, since the `NetworkReachibility` won't work well in the simulator. Either way, the loading view should only depend on whether there is an available local cache or not, not if there is a network available since this could cause unpredictable behaviour.

### 🧪 Testing
N/A. Not really testable in the Demo App, but the customer confirmed the solution works.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
@@ -310,9 +310,7 @@ open class ChatChannelListViewModel: ObservableObject, ChatChannelListController
 
         updateChannels()
 
-        if channels.isEmpty {
-            loading = networkReachability.isNetworkAvailable()
-        }
+        loading = channels.isEmpty
 
         controller?.synchronize { [weak self] error in
             guard let self = self else { return }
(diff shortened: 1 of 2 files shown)


Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
```diff
@@ -310,9 +310,7 @@ open class ChatChannelListViewModel: ObservableObject, ChatChannelListController
 
         updateChannels()
 
-        if channels.isEmpty {
-            loading = networkReachability.isNetworkAvailable()
-        }
```
nuno-vieira, line 315: @martinmitrevski I can probably revert this, since it is not related, but this means that for simulators, the loading view will never appear, and while doing development this can be annoying, especially because most devs initially use the simulator to test things.

Ideally, in the future we should try to follow the Thread List v2 approach, where when there is already a cache, and the data is loading, we show a loading spinner in the header view. As well, as a loading spinner on the footer view when loading more data. (This useful especially in low network conditions)

martinmitrevski: why not update this logic for simulators only? (Keep everything as is for regular devices)

nuno-vieira: That is going to complicate the code IMO. I think we should align this with UIKit, besides that it is usually not a good practice using network reachability for things that can potentially block user interaction 🤔

martinmitrevski (approved):

PR #680, opened 2024-12-05
Fix poor message search performance

nuno-vieira:
### 🔗 Issue Links
Fixes IOS-574

### 🎯 Goal

Fixes the poor performance of Message Search.

### Problem
The root cause of the issue was in the `updateMessageSearchResults()` function. We were accessing the `channelController.channel` for every search result. The problem with this is that we call `startObservers()` for every `channel` property access, which performs a Core Data fetch and, in this case, on the main thread.

We can see this from the time profiler:
image

Besides this, we were calling `updateMessageSearchResults()` twice, for every completion block performed in the `loadNextMessages()` and also in the `didUpdateMessages()` delegate calls.

### First Solution

My first idea was to follow the UIKit approach. There, we use the `dataStore` to fetch the channel directly. However, this also does not work for SwiftUI. The reason is that for UIKit, the `dataStore` is accessed when the cell is reused, whereas in SwiftUI, we map all the results at once, and so we access the `dataStore.channel` for every result again, which means we convert the DTO to model on the main thread multiple times. 

Profile result:
image

So, to avoid converting the channel models every time, I implemented a cache that would reuse the existing channels instead of always doing the DTO -> Model conversion.

This worked and the performance was great, but it does require an additional cache, which can increase the memory and is dependent on the amount of different channels we have in the result. So the final solution (check next section) is probably the better one.

### Final Solution

Eventually, I realised that I could simply use the original implementation but offload the search results mapping to a background thread. This worked flawlessly, and it has the less amount of changes and does not require any additional caching. 

Result (No hangs, and Main thread not busy):
image

### 🧪 Testing

Perform a message search in the Channel List. For example, search for "Hey".

**Known Issue:** The channel name is not rendered for some search results. This is an existing issue and out of the scope of this PR.

### 🎨 Changes

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/ccdf5360-d3d2-4bd1-8185-14928164d0d2" /> | <video src="https://github.com/user-attachments/assets/d4c49534-95ac-46aa-98df-ed65e3458c45" /> |

### ☑️ Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### 🔄 Changed
+- Fix poor message search performance [#680](https://github.com/GetStream/stream-chat-swiftui/pull/680)
 
 # [4.68.0](https://github.com/GetStream/stream-chat-swiftui/releases/tag/4.68.0)
 _December 03, 2024_
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-⚡ Performance
+### ⚡ Performance
 - Improve message search performance [#680](https://github.com/GetStream/stream-chat-swiftui/pull/680)
 
 # [4.68.0](https://github.com/GetStream/stream-chat-swiftui/releases/tag/4.68.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
-- Fix poor message search performance [#680](https://github.com/GetStream/stream-chat-swiftui/pull/680)
+⚡ Performance
+- Improve message search performance [#680](https://github.com/GetStream/stream-chat-swiftui/pull/680)
 
 # [4.68.0](https://github.com/GetStream/stream-chat-swiftui/releases/tag/4.68.0)
 _December 03, 2024_
--- a/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
@@ -105,6 +105,10 @@ open class ChatChannelListViewModel: ObservableObject, ChatChannelListController
     internal var channelListSearchController: ChatChannelListController?
     internal var messageSearchController: ChatMessageSearchController?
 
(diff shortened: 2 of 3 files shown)


Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
```diff
@@ -108,19 +108,3 @@ extension ChatChannel {
...
-            let searchResult = ChannelSelectionInfo(
-                channel: channel,
-                message: self
-            )
-            return searchResult
-        } else {
-            return nil
-        }
-    }
-}
```
nuno-vieira, line 126: Removed the extension because it could be used in other places , and since the performance it not good, better to not make it easy to use.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### 🔄 Changed
+- Fix poor message search performance [#680](https://github.com/GetStream/stream-chat-swiftui/pull/680)
```
martinmitrevski, line 7: "Improve message search performance" sounds better?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
```diff
@@ -423,10 +419,22 @@ open class ChatChannelListViewModel: ObservableObject, ChatChannelListController
             return
         }
 
-        searchResults = messageSearchController.messages
-            .compactMap { message in
-                message.makeChannelSelectionInfo(with: chatClient)
+        queue.async {
+            let results: [ChannelSelectionInfo] = messageSearchController.messages.compactMap { message in
+                guard let channelId = message.cid else { return nil }
+                guard let channel = messageSearchController.client.channelController(for: channelId).channel else {
```
martinmitrevski, line 425: why do we have to go through the message search controller here? We have the client injected in the VM.

martinmitrevski (approved):
LGTM ✅  Left 2 minor comments.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
nuno-vieira: ah yes you are correct 👍

CHANGELOG.md
nuno-vieira: ✅ Done

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
nuno-vieira: Done ✅

laevandus (approved):
With how DB observer work at the moment then this is a way to go.

PR #681, opened 2024-12-05
Fix message thread reply footnote view not shown if parent message not in cache

nuno-vieira:
### 🔗 Issue Link
Fixes https://linear.app/stream/issue/IOS-559/thread-reply-footnote-view-not-shown-in-reply-show-in-channel

### 🎯 Goal
Fixes messages thread reply footnote view not shown when the parent message is not in cache.

#### Problem
The issue is that in order to render the `MessageRepliesView`, we require a `parentMessage`. We get the `parentMessage` from the `dataStore`, but if the parent message is from an old page, it will return `nil` since it won't be in the local DB.

We can't change the `parentMessage` parameter to an `optional` since that would be a breaking change, and changing the Factory protocol could even lead to a silent breaking change for our customers. 

#### Solution

**1st try: Breaking Change**

This one is tricky to solve without breaking changes. First, I tried to force a breaking change by making the function in the Factory @unavailable. But this is not possible. It is only supported by @objc protocols. Since we have default implementations of the factory, creating a real breaking change would be silent for customers, so this is not an option.

**2nd try: Hack-ish Workaround**

The first solution I did without breaking change was to provide the message reply as the parent message as well. Then, check if they had the same ID, and treat it as if parentMessage was nil. You can see the implementation [here](https://github.com/GetStream/stream-chat-swiftui/commit/cfb4b2ed9eeee59c066a54f47577e5c17b52f097).

The problem with this approach is that it is very hacky and unexpected. I customer would need to know about this hack if they wanted to customize the view.

**3rd try: Final Solution** 

The solution I found the best was to create a Lazy view of the existing `MessageRepliesView`. Whenever the parent message does not exist in the local cache, we fetch it and then display the view. From here on, it will be in the local cache until we logout, so it won't be always fetching.

This approach requires the least changes on our side and no changes on the customer side. The con, is whenever this case happens, it means additional requests will be made just to display a footnote view. Ideally, this request would only be executed only when opening the Thread View, but unfortunatly, until we have v5, there's no alternative.

### 🧪 Testing

1. Open a Channel
2. Load old pages
3. Create a Thread Reply and check "Send also in channel"
4. Logout
5. Open the same Channel
6. The "Thread Reply" footnote should be shown in reply created in 3.

### 🎨 Changes

Below is the fix demo. The thread reply view takes a bit of time to load since it is lazy-loaded and requires performing a request to show it. 

| Fix |
|--------|
| <video src="https://github.com/user-attachments/assets/2c8ec23f-e3df-4992-b07b-80fd51d8535c" /> | 

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
@@ -175,12 +175,12 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
                             .accessibility(identifier: "MessageRepliesView")
                         } else if message.showReplyInChannel,
                                   let parentId = message.parentMessageId,
-                                  let controller = utils.channelControllerFactory.currentChannelController {
-                            let parentMessage = controller.dataStore.message(id: parentId) ?? message
+                                  let controller = utils.channelControllerFactory.currentChannelController,
+                                  let parentMessage = controller.dataStore.message(id: parentId) {
                             factory.makeMessageRepliesShownInChannelView(
                                 channel: channel,
                                 message: message,
-                                parentMessage: controller.dataStore.message(id: parentId) ?? message,
+                                parentMessage: parentMessage,
                                 replyCount: parentMessage.replyCount
                             )
                             .accessibilityElement(children: .contain)
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
@@ -175,12 +175,12 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
                             .accessibility(identifier: "MessageRepliesView")
                         } else if message.showReplyInChannel,
                                   let parentId = message.parentMessageId,
-                                  let controller = utils.channelControllerFactory.currentChannelController,
-                                  let parentMessage = controller.dataStore.message(id: parentId) {
+                                  let controller = utils.channelControllerFactory.currentChannelController {
+                            let parentMessage = controller.dataStore.message(id: parentId) ?? message
                             factory.makeMessageRepliesShownInChannelView(
                                 channel: channel,
                                 message: message,
-                                parentMessage: parentMessage,
+                                parentMessage: controller.dataStore.message(id: parentId) ?? message,
                                 replyCount: parentMessage.replyCount
                             )
                             .accessibilityElement(children: .contain)
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
@@ -185,6 +185,20 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
                             )
                             .accessibilityElement(children: .contain)
                             .accessibility(identifier: "MessageRepliesView")
+                        } else if message.showReplyInChannel, let parentId = message.parentMessageId {
+                            /// In case the parent message is not available in the local cache, we need to fetch it from the remote server.
+                            /// The lazy view uses the `factory.makeMessageRepliesShownInChannelView` internally once the parent message is fetched.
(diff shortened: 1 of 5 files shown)


Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -185,6 +185,20 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
...
                             .accessibilityElement(children: .contain)
                             .accessibility(identifier: "MessageRepliesView")
+                        } else if message.showReplyInChannel, let parentId = message.parentMessageId {
+                            /// In case the parent message is not available in the local cache, we need to fetch it from the remote server.
+                            /// The lazy view uses the `factory.makeMessageRepliesShownInChannelView` internally once the parent message is fetched.
+                            LazyMessageRepliesView(
+                                factory: factory,
+                                channel: channel,
+                                message: message,
+                                parentMessageController: chatClient.messageController(
```
martinmitrevski, line 195: Isn't it better to have the message controller as a `@State` in the `LazyMessageRepliesView`? There's a risk of redraws and performing the request several times like this.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageRepliesView.swift
```diff
@@ -116,3 +116,42 @@ public struct MessageRepliesView<Factory: ViewFactory>: View {
         }
     }
 }
+
+/// Lazy view that uses the message controller to fetch the parent message before creating message replies view.
+/// This is need when the parent message is not available in the local cache.
```
martinmitrevski, line 121: *needed

martinmitrevski (commented):
Great idea and solution, it's the best we can do atm ✅ Left one suggestion, let me know what you think.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageRepliesView.swift
```diff
@@ -116,3 +116,42 @@ public struct MessageRepliesView<Factory: ViewFactory>: View {
...
+                    message: message,
+                    parentMessage: parentMessage,
+                    replyCount: parentMessage.replyCount
+                )
+            } else {
+                EmptyView()
+            }
+        }
+    }
+}
```
laevandus, line 157: Did you also explore the way of fetching the missing message(s) in the `ChatChannelDataSource`? Thinking about fetching a page, then checking parentMessageIds and comparing, if the parent message id is part of the currently fetched list of message ids. If not, then fetching it before making the delegate callback. That should take away that delay when the thread reply appears. Could be an alternative, but not sure if better than what you have at the moment (might get messy).

nuno-vieira: IMO the extra complexity is not worth it. Since this is meant to be temporary and although yes it has a bit of delay for the view to appear, this only happens for "Show in channel" messages which is not used very often either way 🤔

martinmitrevski: I agree, not worth to complicate this further.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageRepliesView.swift
```diff
@@ -116,3 +116,42 @@ public struct MessageRepliesView<Factory: ViewFactory>: View {
...
+        factory: Factory,
+        channel: ChatChannel,
+        message: ChatMessage,
+        parentMessageController: ChatMessageController
+    ) {
+        parentMessageObserver = parentMessageController.observableObject
+        self.factory = factory
+        self.channel = channel
+        self.message = message
+        parentMessageObserver.controller.synchronize()
```
martinmitrevski: shouldn't we do this only `onAppear`? `init` can be called many times.

nuno-vieira: Xcode 15 was not happy with the current implementation, so I moved to onAppear, but this requires changing from Group, to use a VStack

martinmitrevski (approved):

PR #687, opened 2024-12-09
[CI] Create a lane to update StreamChat dependency

testableapple:
### 🔗 Issue Link

Resolve https://linear.app/stream/issue/IOS-599

### 🎯 Goal

- Create a lane for bumping the StreamChat dependency

### 🧪 Testing

- Test: https://github.com/GetStream/stream-chat-swiftui/pull/686/files

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -146,6 +146,40 @@ private_lane :appstore_api_key do
   )
 end
 
+desc "Updates StreamChat dependency locally. Usage: `bundle exec fastlane update_stream_chat version:4.56.0`"
+lane :update_stream_chat do |options|
+  raise UI.user_error!('Provide a version.') unless options[:version]
+
+  Dir.chdir('..') do
+    file = 'Package.swift'
+    current_stream_chat_version = File.read(file)[/stream-chat-swift\.git", from: "([\d.]+)"\)/, 1]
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -146,6 +146,40 @@ private_lane :appstore_api_key do
...
+
+    file = 'StreamChatSwiftUI.xcodeproj/project.pbxproj'
+    content = File.read(file)
+    if content.include?("minimumVersion = #{current_stream_chat_version}")
+      File.write(file, content.gsub("minimumVersion = #{current_stream_chat_version}", "minimumVersion = #{options[:version]}"))
+    elsif content.include?('branch = develop')
+      File.write(file, content.gsub('kind = branch', "minimumVersion = #{options[:version]}").gsub('branch = develop', 'kind = upToNextMajorVersion'))
+    else
+      UI.user_error!("Something went wrong after trying to modify #{file}.")
+    end
```
laevandus, line 172: Ideally we should check for stream chat since there are multiple dependencies. Not sure if it is a big deal.

testableapple: Yeah, but it actually makes things much more complex. A chance that we will hit a different dependency with the same key `minimumVersion = #{current_stream_chat_version}` is very low. We also do not use other dependencies from `develop` ever.

laevandus: Yeah, probably fine. We can improve it if really need to. Makes sense to not complicate this too much.

nuno-vieira (approved):
LGTM! ✅

PR #712, opened 2025-01-10
Fix reactions users view not paginating results

nuno-vieira:
### 🔗 Issue Link
Resolves IOS-590

### 🎯 Goal

Fix reactions users view not paginating results.

### 🧪 Testing

1. Open a message with more than 20+ reactions.
2. It should load all reactions.

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/Reactions/ReactionsUsersView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/Reactions/ReactionsUsersView.swift
@@ -80,52 +80,6 @@ struct ReactionsUsersView: View {
     }
 }
 
-class ReactionsUsersViewModel: ObservableObject, ChatMessageControllerDelegate {
-    @Published var reactions: [ChatMessageReaction] = []
-
-    var totalReactionsCount: Int {
-        messageController?.message?.totalReactionsCount ?? 0
-    }
-
-    var isRightAligned: Bool {
-        messageController?.message?.isRightAligned == true
-    }
-
-    private var isLoading = false
-    private let messageController: ChatMessageController?
-    
-    init(message: ChatMessage) {
-        if let cid = message.cid {
-            messageController = InjectedValues[\.chatClient].messageController(
-                cid: cid,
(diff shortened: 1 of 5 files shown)


Sources/StreamChatSwiftUI/ChatChannel/Reactions/ReactionsUsersView.swift
```diff
@@ -70,14 +72,60 @@ struct ReactionsUsersView: View {
...
+            if !viewModel.isRightAligned {
                 Spacer()
             }
         }
         .accessibilityIdentifier("ReactionsUsersView")
     }
 }
 
+class ReactionsUsersViewModel: ObservableObject, ChatMessageControllerDelegate {
+    @Published var reactions: [ChatMessageReaction] = []
```
martinmitrevski, line 84: maybe show the first batch from the latest reactions from the message?

martinmitrevski: since like this we're always making an API call to start pagination, right?

nuno-vieira: No, we are not because the initail data will be already from the latest reactions. So in case there is no more pages, we don't perform the request

Sources/StreamChatSwiftUI/ChatChannel/Reactions/ReactionsUsersView.swift
```diff
@@ -70,14 +72,60 @@ struct ReactionsUsersView: View {
...
-            if !message.isRightAligned {
+            if !viewModel.isRightAligned {
                 Spacer()
             }
         }
         .accessibilityIdentifier("ReactionsUsersView")
     }
 }
 
+class ReactionsUsersViewModel: ObservableObject, ChatMessageControllerDelegate {
```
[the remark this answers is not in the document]
nuno-vieira: ✅

martinmitrevski (approved):

PR #718, opened 2025-01-16
Remove internal _isDatabaseObserverItemReusingEnabled check

laevandus:
### 🎯 Goal

Remove `_isDatabaseObserverItemReusingEnabled` check which has been enabled for 6 months without any side-effects

### 🛠 Implementation

- Remove the check and delete any fallback code
- Remove message caching because it was turned off with that check before
- Deprecate public `userDisplayInfo(from id: String) -> UserDisplayInfo?` method which is not used and only accessed cached data

### 🧪 Testing

N/A

### ☑️ Checklist


nuno-vieira (approved):

martinmitrevski (commented):
Would be great to do some performance checks before merging this.

laevandus:
> Would be great to do some performance checks before merging this.

Ran it with time profiler as well. Nothing shows up related to these changes. That said, I saw some DataStore accesses on the main thread, but this is a separate story.

Issue #723, opened 2025-01-19
Fix: A circle button is visible beneath ChannelAvatarView in ChatChannelView


sheep-q:
## What did you do?
- NavigationBar has its own color, which is different from colors.background (white).
- Using makeChannelHeaderViewModifier of DefaultViewFactory.
- Go to a chat screen from the list chat channel.

## What did you expect to happen?
The channel avatar view should show correctly like this: 
![Image](https://github.com/user-attachments/assets/3ae11824-912d-4523-8e46-3693f083b753)

## What happened instead?
A circle button, which is filled colors.background, is visible beneath ChannelAvatarView in ChatChannelView:
![Image](https://github.com/user-attachments/assets/08b60a32-9e33-465f-afb2-06d6896ee65f)

## Fix:
Configure style for button in ToolbarItem, which automates adding padding for Button, it seems like a bug of SwiftUI.
-> Add this line `.buttonStyle(.plain)` to [here](https://github.com/GetStream/stream-chat-swiftui/blob/1effcd58f2982653f708c732e1e0f9ab94c5817b/Sources/StreamChatSwiftUI/ChatChannel/ChannelHeader/ChatChannelHeaderViewModifier.swift#L73) like this: 
<img width="613" alt="Image" src="https://github.com/user-attachments/assets/e8e2b22f-c5b8-4cef-80ca-8141a053f200" />

## GetStream Environment
**GetStream Chat version:** 4.66.0
**GetStream Chat frameworks:** StreamChat, StreamChatSwiftUI
**iOS version:** iOS 17, iOS 18
**Swift version:** 5.5 
**Xcode version:** 16.2
**Device:** both device + simulator

## Additional context
Condition to reproduce: NavigationBar has its own color, which is different from colors.background (white).

laevandus:
Hi @sheep-q, 
Thank you for reporting this issue. Since you already identified the issue and the solution, would you be interested in making a contribution to the SDK? We are open to contributions from everyone. If you prefer for us to take care of this issue then please let me know and I'll add it to our backlog for review.

Best,
Toomas

sheep-q:
Hi @laevandus. Glad to hear that. Let me make a contribution to the SDK.

sheep-q:
Hi @laevandus, I've created [a PR](https://github.com/GetStream/stream-chat-swiftui/pull/725) for this issue, looking forward to hearing from you.

laevandus:
[...]

PR #725, opened 2025-01-21
Fix white circle button is visible beneath the ChannelAvatarView

sheep-q:
### 🔗 Issue Link
[Github issue](https://github.com/GetStream/stream-chat-swiftui/issues/723)

### 🎯 Goal
The channel avatar view should show correctly without a visible white circle below.

### 🛠 Implementation
- Change the style of button inside ToolbarItem
- Update the initialization of ChannelAvatarView to remove deprecated init warninig.

### 🎨 Changes

image

### ☑️ Checklist

- [X] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/ChannelHeader/ChatChannelHeaderViewModifier.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/ChannelHeader/ChatChannelHeaderViewModifier.swift
@@ -64,12 +64,12 @@ public struct DefaultChatChannelHeader: ToolbarContent {
                     resignFirstResponder()
                     isActive = true
                 } label: {
-                    Rectangle()
-                        .fill(Color(colors.background))
-                        .contentShape(Rectangle())
-                        .frame(width: 36, height: 36)
-                        .clipShape(Circle())
-                        .offset(x: 8)
+                    ChannelAvatarView(
+                        avatar: headerImage,
+                        showOnlineIndicator: onlineIndicatorShown,
+                        size: CGSize(width: 36, height: 36)
+                    )
+                    .offset(x: 4)
                 }
                 .accessibilityLabel(Text(L10n.Channel.Header.Info.title))
 
@@ -78,15 +78,6 @@ public struct DefaultChatChannelHeader: ToolbarContent {
                 } label: {
                     EmptyView()
                 }
-                .accessibilityHidden(true)
-                
-                ChannelAvatarView(
-                    avatar: headerImage,
-                    showOnlineIndicator: onlineIndicatorShown,
-                    size: CGSize(width: 36, height: 36)
-                )
-                .offset(x: 8)
-                .allowsHitTesting(false)
                 .accessibilityHidden(true)
             }
             .accessibilityIdentifier("ChannelAvatarView")


Sources/StreamChatSwiftUI/ChatChannel/ChannelHeader/ChatChannelHeaderViewModifier.swift
```diff
@@ -35,16 +35,13 @@ public struct DefaultChatChannelHeader: ToolbarContent {
     }
 
     public var channel: ChatChannel
-    public var headerImage: UIImage
```
laevandus, line 38: Hi, we can't do any changes to the public API at the moment (only in major versions, but the next one is a minor upgrade). Could you please take these out and just keep the buttonStyle change.
Is there an additional issue with the `headerImage`?

sheep-q: Got it, I deleted headerImage because when using it in the initialization of ChannelAvatarView, it shows warning `init(avatar:showOnlineIndicator:size:)' is deprecated: Use automatically refreshing avatar initializer`.
But no problem, it's not the main reason so I just keep the buttonStyle change

Sources/StreamChatSwiftUI/ChatChannel/ChannelHeader/ChatChannelHeaderViewModifier.swift
```diff
@@ -71,6 +71,7 @@ public struct DefaultChatChannelHeader: ToolbarContent {
                         .clipShape(Circle())
                         .offset(x: 8)
                 }
+                .buttonStyle(.plain)
```
laevandus, line 74: I did some testing and seems like it also changes spacings:
Before:
Before
After:
After
I'll investigate this as well.

laevandus: What does not break the padding is changing the Button's offset from 8 -> 4.

sheep-q: I think changing the Button's offset causes confusing. Clearly, this offset should be equal offset of `ChannelAvatarView`, either both set to 0 or 8. 
I believe that, the issue was from SwiftUI framework that automatically add padding to `ToolbarItem(placement: .navigationBarTrailing)`. I tested the same ZStack's content to several wrapped views without encountering any issues

sheep-q: Or you can just configure Button like this without multiple layers in ZStack, it doesn't care about additional padding.
I'm not sure about the purpose of your using separate layers, so I wouldn't use this approach.
```swift
               Button {
                    resignFirstResponder()
                    isActive = true
                } label: {
                    ChannelAvatarView(
                        avatar: headerImage,
                        showOnlineIndicator: onlineIndicatorShown,
                        size: CGSize(width: 36, height: 36)
                    )
                    .offset(x: 8)
                }
```

laevandus: This change also has to work for iOS 14 which has brought most of the complexities.

laevandus: Hi @sheep-q I did some testing on older iOS versions and this is OK. Also looked into the question of the offset, and the correct is actually 4, not 8. Please change your snippet above to use 4 and let's get this change in.

sheep-q: Hi @laevandus, Would you like to change the offset for both Button and ChannelAvatarView or just for Button

sheep-q: which solution did you approve? 
[Wrap the ChannelAvatarView inside the Button](https://github.com/GetStream/stream-chat-swiftui/pull/725#discussion_r1923491494) or modify the Button style (current change)

sheep-q: Hi @laevandus, I've pushed an updated commit, take a look. thank you

laevandus: Thank you @sheep-q I'll take it in soon.

laevandus (approved):
Thank you for your contribution! ✅

laevandus:
I will add a CHANGELOG entry in a separate PR. Also verified that this change does not create any snapshot test failures.

PR #734, opened 2025-01-29
Customize channel avatar view

martinmitrevski:
### 🔗 Issue Link
https://linear.app/stream/issue/IOS-684/[swiftui]-make-channel-avatar-view-customizable

### 🎯 Goal

At the moment, there's no possibility to customize only the channel avatar view. If you want to change it, you need to implement several factory methods, all with bigger views involved.

### 🛠 Implementation

Exposed a new factory method.

### 🧪 Testing

Added snapshot tests for the customization, everything else should work the same as before.

### 🎨 Changes

### ☑️ Checklist


Sources/StreamChatSwiftUI/ChatChannel/ChannelHeader/ChatChannelHeaderViewModifier.swift
```diff
@@ -11,7 +11,7 @@ public protocol ChatChannelHeaderViewModifier: ViewModifier {
 }
 
 /// The default channel header.
-public struct DefaultChatChannelHeader: ToolbarContent {
+public struct DefaultChatChannelHeader<Factory: ViewFactory>: ToolbarContent {
```
nuno-vieira, line 14: Isn't this a breaking change?

Sources/StreamChatSwiftUI/DefaultViewFactory.swift
```diff
@@ -95,6 +96,18 @@ extension ViewFactory {
         )
     }
     
+    public func makeChannelAvatarView(
+        for channel: ChatChannel,
+        showOnlineIndicator: Bool,
+        size: CGSize
```
nuno-vieira, line 102: It will be easy to break this in the future. There's is a chance we might introduce other stuff. Better to create a `ChannelAvatarViewOptions` struct, something like:

```swift
struct ChannelAvatarViewOptions {
    var showOnlineIndicator: Bool
    var size: CGSize
}
```

Then:
```swift
public func makeChannelAvatarView(
    for channel: ChatChannel,
    with options: ChannelAvatarViewOptions
)
```

We have something similar on UIKit

Sources/StreamChatSwiftUI/ViewFactory.swift
```diff
@@ -58,6 +58,19 @@ public protocol ViewFactory: AnyObject {
         trailingSwipeLeftButtonTapped: @escaping (ChatChannel) -> Void,
         leadingSwipeButtonTapped: @escaping (ChatChannel) -> Void
     ) -> ChannelListItemType
+    
+    associatedtype ChannelAvatarViewType: View
+    /// Creates the channel avatar view shown in the channel list, search results and the channel header.
+    /// - Parameters:
+    /// - channel: the channel where the avatar is displayed.
+    /// - showOnlineIndicator: whether the online indicator is shown.
+    /// - size: the size of the avatar.
```
nuno-vieira, line 67: As discussed above, would be nice to introduce an Options struct, so that it is more future proof

Sources/StreamChatSwiftUI/DefaultViewFactory.swift
martinmitrevski: good point, let me do that

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
```diff
@@ -127,6 +130,16 @@ public struct ChatChannelListItem: View {
...
 
+/// Options for setting up the channel avatar view.
+public struct ChannelAvatarViewOptions {
+    /// Whether the online indicator should be shown.
+    public var showOnlineIndicator: Bool
+    /// Size of the avatar.
+    public var size: CGSize = .defaultAvatarSize
+    /// Optional avatar image. If not provided, it will be loaded by the channel header loader.
+    public var avatar: UIImage?
+}
```
nuno-vieira, line 141: I think we need to make the init public as well right? 🤔

nuno-vieira (approved):
LGTM! ✅

We just need to provide the missing public init and fix the tests 👍

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
martinmitrevski: as discussed, we're keeping it internal for now.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
```diff
@@ -157,9 +170,10 @@ public struct ChannelAvatarView: View {
     public init(
         channel: ChatChannel,
         showOnlineIndicator: Bool,
+        avatar: UIImage? = nil,
         size: CGSize = .defaultAvatarSize
     ) {
-        avatar = nil
+        self.avatar = avatar
```
laevandus, line 176: We should not load the channel avatar if this is set. `reloadAvatar` should have a `avatar == nil` check

martinmitrevski: you mean sth like this?
```swift
private func reloadAvatar() {
        guard let channel, avatar == nil else { return }
        channelAvatar = utils.channelHeaderLoader.image(for: channel)
}
```

laevandus: Yep

martinmitrevski: done

laevandus (approved):
✅

Sources/StreamChatSwiftUI/ViewFactory.swift
```diff
@@ -58,6 +58,17 @@ public protocol ViewFactory: AnyObject {
...
         leadingSwipeButtonTapped: @escaping (ChatChannel) -> Void
     ) -> ChannelListItemType
+    
+    associatedtype ChannelAvatarViewType: View
+    /// Creates the channel avatar view shown in the channel list, search results and the channel header.
+    /// - Parameters:
+    ///  - channel: the channel where the avatar is displayed.
+    ///  - options: the options used to configure the avatar view.
+    /// - Returns: view displayed in the channel avatar slot.
+    func makeChannelAvatarView(
```
martinmitrevski, line 68: @ipavlidakis fyi

PR #754, opened 2025-02-12
Add support for pasting images into composer text input

anh-ngquang:
### 🔗 Issue Link
N/A

### 🎯 Goal

Allow pasting images into composer text input, which is a common feature in messaging apps.

### 🛠 Implementation

Allow paste action on `InputTextView`, and call a closure named `onImagePasted` to pass the pasted image up to the `MessageComposerView`, then handle the image with the `imagePasted` method on `MessageComposerViewModel`, which save the image to a local temporary file, create a new `AddedAsset` and insert it to the `addedAssets` array.

### 🧪 Testing

Copy an image into the clipboard and paste into message composer text input. The image should be added into the composer view the same way as choosing an image from Photo library.

### 🎨 Changes

https://github.com/user-attachments/assets/41c4470d-87cf-4a70-b6bb-27e1ba8a5ea3

### ☑️ Checklist


Sources/StreamChatSwiftUI/ViewFactory.swift
```diff
@@ -637,7 +637,8 @@ public protocol ViewFactory: AnyObject {
         placeholder: String,
         editable: Bool,
         maxMessageLength: Int?,
-        currentHeight: CGFloat
+        currentHeight: CGFloat,
+        onImagePasted: ((UIImage) -> Void)?
```
nuno-vieira, line 641: @anh-ngquang Thanks for the PR 👍 The implementation will probably require some changes because this is a breaking change. We will take it from here once we have time. We are planning on working on this in the next sprint.

nuno-vieira:
New PR without breaking changes available here: https://github.com/GetStream/stream-chat-swiftui/pull/797

Issue #759, opened 2025-02-18
HEIC images return incorrect MIME Type (application/octet-stream)


sheep-q:
## What did you do?
- Send a HEIC image - High Efficiency Image Container (HEIC) format, which Apple uses on iPhones and iPads
- Check type and mimeType of image from `AttachmentFileType`

## What did you expect to happen?
- Expect correct file type:`heic` and mime type: `image/heic`

## What happened instead?
HEIC images return incorrect data with:
- type: `generic`
- mime type: `application/octet-stream`

## GetStream Environment
**GetStream Chat version:** 4.72.0
**GetStream Chat frameworks:** StreamChat, StreamChatSwiftUI
**iOS version:** iOS 17.5
**Swift version:** Swift 5
**Xcode version:** Xcode 16.2
**Device:** iPhone 15 pro

## Additional context
My approach to address the issue: Add `heic` type and `image/heic` mime type to `AttachmentFileType` Enum

PR #767, opened 2025-02-28
Upload photos from the library as JPEG instead of HEIC

nuno-vieira:
### 🔗 Issue Link
https://github.com/GetStream/stream-chat-swiftui/issues/759

### 🎯 Goal

Send photos from the library as JPEG instead of HEIC to make sure other platforms can show the images correctly. 

### 🛠 Implementation
We now convert the HEIC image to JPG when adding it to the composer.

This approach is what we do when sending photos from the Camera. With this change, we are also consistent with UIKit as well which follows the same approach.

### 🧪 Testing

Smoke testing around images

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/Composer/PhotoAttachmentPickerView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/Composer/PhotoAttachmentPickerView.swift
@@ -102,7 +102,7 @@ public struct PhotoAttachmentCell: View {
                             .allowsHitTesting(true)
                             .onTapGesture {
                                 withAnimation {
-                                    if let assetURL = assetURL {
+                                    if let assetURL = assetJpgURL() {
                                         onImageTap(
                                             AddedAsset(
                                                 image: image,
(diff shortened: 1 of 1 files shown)


Sources/StreamChatSwiftUI/ChatChannel/Composer/PhotoAttachmentPickerView.swift
```diff
@@ -193,4 +193,13 @@ public struct PhotoAttachmentCell: View {
...
         }
     }
+
+    /// The original photo is usually in HEIC format.
+    /// This makes sure that the photo is converted to JPG.
+    /// This way it is more compatible with other platforms.
+    private func assetJpgURL() -> URL? {
+        guard let assetURL = assetURL else { return nil }
+        guard let assetData = try? Data(contentsOf: assetURL) else { return nil }
+        return try? UIImage(data: assetData)?.temporaryLocalFileUrl()
```
laevandus, line 203: Not for this PR, but I did not know that `temporaryLocalFileUrl` actually goes and converts the image to JPEG. We should probably rename this at some point.

nuno-vieira: Yeah I actually change it and then reverted it. But I think we can do it already here. Although on UIKit we have the same function

nuno-vieira: > I guess this is fast enough that we don't need any async call for this.

Yes it is quite fast, and is also only executed when an attachment is added to the composer

nuno-vieira: Done ✅ Let me know WDYT

PR #771, opened 2025-03-04
[CI] Bump Xcode, macOS and iOS versions

testableapple:
### 🔗 Issue Link

Resolve https://linear.app/stream/issue/IOS-683

### Test

https://github.com/GetStream/stream-chat-swiftui/actions/runs/13657080268


laevandus (approved):
Looks good! ✅ Seems like we had some snapshots lingering around.

PR #776, opened 2025-03-07
Show translated text for participant messages

laevandus:
### 🔗 Issue Link

Resolves: [IOS-724](https://linear.app/stream/issue/IOS-724/)

### 🎯 Goal

Support showing auto-translated text

### 🛠 Implementation

> [!NOTE]
> Added debug view support since otherwise it is hard to manually test this.

When [auto-translation](https://getstream.io/chat/docs/ios-swift/translation/#enabling-automatic-translation) is enabled, we get back translated text with the message payload. `ChatMessage.text` is always the original text.  

Added the same "Translated to XYZ" label to messages which are translated (like UIKit). Use channel's membership for accessing the preferred language (like UIKit).

### 🧪 Testing

1. Launch the app, A uses app configuration and chooses a language, let's say `es`
2. Launch the app on a second simulator, B uses app configuration and chooses a language, let's say `en`
3. A creates a channel with B
4. A sends message `Hola`, B sees it as `Hello`
5. B sends `Hello`, A sees it as `Hola`

### 🎨 Changes

https://github.com/user-attachments/assets/e7e7be8b-e67f-473e-9e85-d45a7a95b7aa

### ☑️ Checklist


Sources/StreamChatSwiftUI/Utils/MessageCachingUtils.swift
```diff
@@ -19,9 +19,22 @@ class MessageCachingUtils {
...
+    // ChatMessage.adjustedText is used everywhere, but there is no easy way to get the
+    // language used in the channel. Therefore, it is cached here.
+    func translationLanguage(for cid: ChannelId) -> TranslationLanguage? {
+        if let language = channelTranslationLanguages[cid] {
+            return language
+        }
+        let language = InjectedValues[\.chatClient].channelController(for: cid).channel?.membership?.language
+        channelTranslationLanguages[cid] = language
+        return language
+    }
```
laevandus, line 33: Struggled to think about a good solution here. UIKit's implementation uses `channel.membership.language` for figuring out which language to show if there are translations. `ChatMessage.adjustText` has no direct access to get that information. 
I am not really happy about how this looks like at the moment, but can't figure out a better way. Ideally, ChatMessage would have some sort of property which tells the language to use for translations. That would be must cleaner, but would involve adding a new property to ChatMessage. When ChatMessage is created, we could access channel's `membership.language` and read the state from there. In summary, we would add `ChatMessage.translationLanguage` or something similar. WDYT?

laevandus: We can have a discussion in the next meeting on Monday if needed.

laevandus: I'll give it another try. This also needs to work in the channel list so I need to make sure it works there as well. I'll see if it would be better to try to explicitly handle it and not doing it in the adjustText method where is no channel information available.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListView.swift
```diff
@@ -597,3 +598,18 @@ private class MessageRenderingUtil {
...
+extension EnvironmentValues {
+    var channelTranslationLanguage: TranslationLanguage? {
+        get {
+            self[ChannelTranslationLanguageKey.self]
+        }
+        set {
+            self[ChannelTranslationLanguageKey.self] = newValue
+        }
+    }
+}
```
laevandus, line 615: Channel data is not available in the message's text view. Interfaces are public so no easy way to forward this information with function arguments. Next best way is to use environment key for this.

Note: `MessageContainerView` is the last one to have access to channel.

Sources/StreamChatSwiftUI/Utils/MessageCachingUtils.swift
laevandus: I think I found a better way. It sure got too hacky with trying to do this in the `adjustedText`.

Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerView.swift
```diff
@@ -300,6 +300,7 @@ public struct ComposerInputView<Factory: ViewFactory>: View, KeyboardReadable {
                     isInComposer: true,
                     scrolledId: .constant(nil)
                 )
+                .environment(\.channelTranslationLanguage, viewModel.channelController.channel?.membership?.language)
```
martinmitrevski, line 303: Customers might have their own message composer view implementation. Just to make sure, if we access it `@Environment(\.channelTranslationLanguage) var translationLanguage`, but this line is not called, it won't crash, because it's optional, right?

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListView.swift
martinmitrevski: V5 is a good option to improve this.

martinmitrevski (approved):
this also works, just check that it won't crash under circumstances where customers might replace views where we setup the environment.

nuno-vieira (approved):
LGTM! ✅

Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerView.swift
laevandus: Yes, then it uses the the default value for the key which is nil

testableapple (approved):

PR #779, opened 2025-03-10
Make sendMessage in MessageComposerViewModel open

martinmitrevski:
### 🔗 Issue Link
Resolves https://linear.app/stream/issue/IOS-732/provide-message-sending-params-to-messagecomposerview.

### 🎯 Goal

### 🛠 Implementation

_Provide a description of the implementation._

### 🧪 Testing

_Describe the steps how this change can be tested (or why it can't be tested)._

### 🎨 Changes

### ☑️ Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerView.swift
@@ -22,6 +22,7 @@ public struct MessageComposerView<Factory: ViewFactory>: View, KeyboardReadable 
     @Binding var editedMessage: ChatMessage?
     
     private let recordingViewHeight: CGFloat = 80
+    private let messageSendingParameters: MessageSendingParameters
 
     public init(
         viewFactory: Factory,
@@ -30,6 +31,7 @@ public struct MessageComposerView<Factory: ViewFactory>: View, KeyboardReadable 
         messageController: ChatMessageController? = nil,
         quotedMessage: Binding<ChatMessage?>,
         editedMessage: Binding<ChatMessage?>,
+        messageSendingParameters: MessageSendingParameters = .init(),
         onMessageSent: @escaping () -> Void
     ) {
         factory = viewFactory
@@ -43,6 +45,7 @@ public struct MessageComposerView<Factory: ViewFactory>: View, KeyboardReadable 
         _quotedMessage = quotedMessage
         _editedMessage = editedMessage
         self.onMessageSent = onMessageSent
+        self.messageSendingParameters = messageSendingParameters
     }
 
     @StateObject var viewModel: MessageComposerViewModel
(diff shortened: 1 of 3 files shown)


Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerView.swift
```diff
@@ -22,6 +22,7 @@ public struct MessageComposerView<Factory: ViewFactory>: View, KeyboardReadable
     @Binding var editedMessage: ChatMessage?
     
     private let recordingViewHeight: CGFloat = 80
+    private let messageSendingParameters: MessageSendingParameters
```
nuno-vieira, line 25: The problem I see with this approach is that these parameters will be static, right? The customer set's it once, but then it can't change it dynamically. So if you set `skipPush: true`, then all messages will have skipPush to `true` right? 

Other alternatives I'm thinking:
- Making the `ViewModel.sendMessage()` open (This way the customer can override the function and change the parameters before calling the ` super.sendMessage()`
- Make `messageSendingParameters` a `var` and not `let`, and/or maybe provide a closure to change the paremeters before calling the sendMessage() 🤔

nuno-vieira: The first one looks like it should be more than enough, though. Unless I'm missing something.

martinmitrevski: subclassing won't help - customers don't want to customize the send message method, but pass different params from the view layer. Even if we make it open (which we can for sure), they would still need to re-implement the composer view.
A different option I considered is maybe expose the whole closure and let customers handle it all by themselves, basically this part can be `nil` by default with this default implementation:
```
                   viewModel.sendMessage(
                        quotedMessage: quotedMessage,
                        editedMessage: editedMessage,
                        isSilent: messageSendingParameters.isSilent,
                        skipPush: messageSendingParameters.skipPush,
                        skipEnrichUrl: messageSendingParameters.skipEnrichUrl,
                        extraData: messageSendingParameters.extraData
                    ) {
                        quotedMessage = nil
                        editedMessage = nil
                        onMessageSent()
                    }
```

nuno-vieira: By subclassing, they can change the paremeters as well, without changing the existing implementation. Example:

```swift
open class CustomComposerViewModel: MessageComposerViewModel {

    public var customShouldSkip: Bool? = false

    open override func sendMessage(
        quotedMessage: ChatMessage?,
        editedMessage: ChatMessage?,
        isSilent: Bool = false,
        skipPush: Bool = false,
        skipEnrichUrl: Bool = false,
        extraData: [String : RawJSON] = [:],
        completion: @escaping () -> Void
    ) {
        super.sendMessage(
            quotedMessage: quotedMessage,
            editedMessage: editedMessage,
            isSilent: isSilent,
            skipPush: customShouldSkip ?? skipPush,
            skipEnrichUrl: skipEnrichUrl,
            extraData: extraData,
            completion: completion
        )
    }
}
``` 

Why is this not enough?

martinmitrevski: ah yes, you are totally right! They can pass the custom VM instead. Let me change stuff.

martinmitrevski: done

nuno-vieira (approved):
LGTM! ✅

laevandus (approved):
✅

PR #582, opened 2024-08-15
Fix updating of mentioned users when sending a message

martinmitrevski:
### 🔗 Issue Link
Resolves https://stream-io.atlassian.net/browse/PBE-5569.

### 🎯 Goal

Fix updating of mentioned users.

### 🛠 Implementation

_Provide a description of the implementation._

### 🧪 Testing

Comment out `makeMessageViewModifier` in the `DemoAppFactory`. Send a message that mentions a user. It should highlight the user.

### 🎨 Changes

_Add relevant screenshots or videos showcasing the changes._

### ☑️ Checklist

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Affected documentation updated (docusaurus, tutorial, CMS (task created)

PR #815, opened 2025-04-24
Introduce `MessageViewModel` + Show original translated message

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-794/show-original-translation

### 🎯 Goal

- Introduce a new MessageViewModel so it is easier to customise the message view logic.
- Add support for showing the original translated message text

#### Goal
The main goal of this PR is to introduce the new `MessageViewModel` which was required to properly implement the original translated message feature without breaking changes or too many hacks. Besides this, a lot of customers want to override the logic of the Message View, especially when to hide or show some views based on business logic. 

#### Requirement
One thing that was important to do was to have a way to share state from the `ChatChannelViewModel` to the new `MessageViewModel`, since the `MessageViewModel` should not have any internal state to not cause unnecessary re-renders. Besides that, since the message view is inside a LazyStack, the state would be overidden whenever the view is recreated. So it is important to provide the message view model to each message view whenever it is created.

#### Solution
The most common practice is to have item view models as the data of the `ListViewModel`. So instead of `messages: [ChatMessage]` as the data, it would be `messages: [MessageViewModel]`. But because of our `LazyCachedMapCollection`, and to avoid breaking changes, we can't really do this. So the solution was to introduce a factory method in the `ChatChannelViewModel`, the `makeMessageViewModel(message:)`. This way, we can pass data to the message view model from the channel view model, and the view model is always updated whenever the view is also re-created. The view model is then passed to the `MessageContainerView` as an environment object. (Can't be @ObservedObject, otherwise it would be breaking)

### 🧪 Manual Testing Notes

TODO

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -7,6 +7,8 @@ import StreamChat
 import SwiftUI
 
 public struct MessageContainerView<Factory: ViewFactory>: View {
+    @EnvironmentObject var channelViewModel: ChatChannelViewModel
+    @EnvironmentObject var messageViewModel: MessageViewModel
```
nuno-vieira, line 11: I have a stashed commit that uses `@EnvinronmentValues` instead, like we do with the `@Environment(\.channelTranslationLanguage)`. Ofc, there are pros and cons here:

**@EnvironmentObject:**
+ No Optionals
- Unsafe, can cause crashes easily if a customer is using views directly and do not provide the env object

**@Environment:**
+ Safer
- More boilerplate code, to fallback to previous logic when the view model is optional

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -231,10 +227,30 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
...
+                            Button(
+                                action: {
+                                    if messageViewModel.originalTextShown {
+                                        channelViewModel.showTranslatedText(for: message)
+                                    } else {
+                                        channelViewModel.showOriginalText(for: message)
+                                    }
+                                },
+                                label: {
+                                    Text(messageViewModel.originalTextShown ? "Show Translation" : "Show Original")
```
nuno-vieira, line 246: Still requires adding L10n translation. And the UI is not fully done yet.

StreamChatSwiftUITests/Tests/ChatChannel/MessageListViewAvatars_Tests.swift
```diff
@@ -101,7 +105,7 @@ class MessageListViewAvatars_Tests: StreamChatTestCase {
             onScrollToBottom: {},
             onLongPress: { _ in }
         )
-
-        return messageListView
+        .applyDefaultSize()
+        .environmentObject(ChatChannelViewModel(channelController: channelController))
```
nuno-vieira, line 109: Add to update a lot of tests, and there is still more tests to update because of the `@EnvironmentObject`. So another advantage of `@Environment` is that it won't be required to update the tests since it will fallback to the previous implementation.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
```diff
@@ -252,7 +254,8 @@ struct StreamTextView: View {
 public struct LinkDetectionTextView: View {
     @Environment(\.layoutDirection) var layoutDirection
     @Environment(\.channelTranslationLanguage) var translationLanguage
-    
+    @EnvironmentObject private var viewModel: MessageViewModel
```
nuno-vieira, line 257: This one is a good example that @EnvironmentObject is not a good idea here, since `LinkDetectionTextView` is likely to be reused as a standalone by other customers.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
```diff
@@ -275,30 +276,16 @@ public struct LinkDetectionTextView: View {
...
         }
         .foregroundColor(textColor(for: message))
         .font(fonts.body)
         .tint(tintColor)
-        .onAppear {
-            displayedText = attributedString(for: message)
-        }
-        .onChange(of: message, perform: { updated in
-            displayedText = attributedString(for: updated)
-        })
```
nuno-vieira, line 292: I will probably need some help here CC @martinmitrevski  @laevandus. This was overriding the view model content always, and was causing some trouble to me. Why is this needed?

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
```diff
@@ -275,30 +276,16 @@ public struct LinkDetectionTextView: View {
...
     
-    private func attributedString(for message: ChatMessage) -> AttributedString {
-        var text = message.adjustedText
-        
-        // Translation
-        if let translatedText = message.textContent(for: translationLanguage) {
-            text = translatedText
-        }
+    var displayText: AttributedString {
+        let text = viewModel.textContent
```
nuno-vieira, line 287: Having just this was enough to make it work, and I tried markdown etc and did not break anything for me.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
```diff
@@ -275,30 +276,16 @@ public struct LinkDetectionTextView: View {
     
     public var body: some View {
         Group {
-            if let displayedText {
-                Text(displayedText)
-            } else {
-                Text(message.adjustedText)
-            }
```
nuno-vieira, line 282: In case of `LinkDetectionTextView`, I'm considering passing a new property, maybe `originalText: String?` instead of the whole view model, so that it requires less changes. Still need to check if it would work.

martinmitrevski (commented):
I think there's a simpler way, without using environment object and involving the channel view model.
- we pass the message view model in the MessageContainerView, with default value. We can do that, it's not a breaking change.
- we introduce a state there for each message, whether the show original is tapped for that message.
- customers will implement the existing makeMessageContainer method, by passing their own VM to the regular view.
Let's discuss it when you are back, but I think we can simplify things here a bit.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
laevandus: Why not @StateObject although then we need to change inits.

laevandus: Feels like we should go for @StateObject even when it requires making some inits deprecated. Maybe it is time to create a new text view instead? RichTextView or something since the naming does not reflect anymore what it does (does much more than links). New type would allow making bigger changes.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListView.swift
```diff
@@ -603,6 +608,14 @@ private struct ChannelTranslationLanguageKey: EnvironmentKey {
     static let defaultValue: TranslationLanguage? = nil
 }
 
+private struct MessageViewModelKey: EnvironmentKey {
+    static let defaultValue: MessageViewModel? = nil
+}
+
+private struct ChatChannelViewModelKey: EnvironmentKey {
+    static let defaultValue: ChatChannelViewModel? = nil
+}
```
laevandus, line 617: These seems to be unused now

nuno-vieira: It is used on LinkDetectionView, and it would be used for lower level views if needed

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -265,43 +264,42 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
...
+        .background(messageViewModel.isPinned ? Color(colors.pinnedBackground) : nil)
+        .padding(.bottom, messageViewModel.isPinned ? paddingValue / 2 : 0)
         .transition(
             message.isSentByCurrentUser ?
                 messageListConfig.messageDisplayOptions.currentUserMessageTransition :
                 messageListConfig.messageDisplayOptions.otherUserMessageTransition
         )
         .accessibilityElement(children: .contain)
         .accessibilityIdentifier("MessageContainerView")
+        .id(message.id + messageViewModel.textContent)
```
nuno-vieira, line 289: This was added so that whenever the view model textContent changes, it should reset the view. This is useful for translations, markdown etc..

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
```diff
@@ -0,0 +1,152 @@
...
+    ) {
+        self.message = message
+        self.channel = channel
+        self.originalTranslationsStore = originalTranslationsStore
+        self.originalTranslationsStore.$originalTextMessageIds.sink(
+            receiveValue: { [weak self] _ in
+                self?.objectWillChange.send()
+            }
+        )
+        .store(in: &cancellables)
```
nuno-vieira, line 31: Since this is a nested observable object, we need to notify the parent object that the translations have changed

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
martinmitrevski: hmm, this doesn't feel right. The view should re-render automatically when the text changes.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    private var messageListConfig: MessageListConfig {
+        utils.messageListConfig
+    }
+}
+
+/// A singleton store that keeps track of which messages have their original text shown.
+///
+/// **Note:** This is not thread-safe, it should only be used on the main thread.
+public class MessageOriginalTranslationsStore: ObservableObject {
```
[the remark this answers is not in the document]
martinmitrevski: You can put it on the main actor already, no need to wait for Swift 6 migration

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
```diff
@@ -263,8 +264,6 @@ public struct LinkDetectionTextView: View {
         LocalizedStringKey(message.adjustedText)
     }
     
-    @State var displayedText: AttributedString?
```
martinmitrevski, line 266: We should be careful here of breaking changes around markdown links, will try to find the threads so we can test.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -271,43 +270,42 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
...
+        .environment(\.messageViewModel, messageViewModel)
```
martinmitrevski, line 296: this is needed only for the link detector text view, right? Maybe we add V5 todo?

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
martinmitrevski: why do you need this? Can't you just make the store `@Published`?

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
nuno-vieira: Without this, it does not work, tho, the textContent is not updated. I think it is most likely due to how we have the LinkDetectionTextView

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
nuno-vieira: No it wont work, because this is nested observable object, even with @Published

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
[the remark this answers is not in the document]
nuno-vieira: its gives a couple of warning tho, but I guess it is fine

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
```diff
@@ -0,0 +1,152 @@
...
+    private init() {}
+
+    public static let shared = MessageOriginalTranslationsStore()
```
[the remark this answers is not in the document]
nuno-vieira: Ah yes, I forgot about this. I think what we can do is, clear the store when the channel is open or closed. In this case the state will only be per channel.

Sources/StreamChatSwiftUI/ChatChannel/Reactions/ReactionsOverlayView.swift
```diff
@@ -225,6 +233,7 @@ public struct ReactionsOverlayView<Factory: ViewFactory>: View {
             isFirst: messageDisplayInfo.isFirst,
             scrolledId: .constant(nil)
         )
+        .environment(\.messageViewModel, messageViewModel)
```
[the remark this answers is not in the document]
nuno-vieira: We need this for LinkDetectionTextView to show the correct text in the Reaction Overlay view

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
martinmitrevski: This one is important to fix, since the explicit view being setup like this might prevent other updates (adding reactions, read counts etc). So try to figure it out, if you don't have luck, I will also have a go on this one.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
martinmitrevski: is that still an issue? It's there from iOS 13 😅

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
nuno-vieira: Discussed internally and now it is fixed 👍

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
nuno-vieira: Yeah, I think only new @Observable macro fixes this

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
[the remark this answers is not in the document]
nuno-vieira: Not warnings actually, it does not compile at all. So better to wait for Swift 6

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageListConfig.swift
```diff
@@ -171,7 +172,8 @@ public struct MessageDisplayOptions {
             .defaultLinkDisplay,
         spacerWidth: @escaping (CGFloat) -> CGFloat = MessageDisplayOptions.defaultSpacerWidth,
         reactionsTopPadding: @escaping (ChatMessage) -> CGFloat = MessageDisplayOptions.defaultReactionsTopPadding,
-        dateSeparator: @escaping (ChatMessage, ChatMessage) -> Date? = MessageDisplayOptions.defaultDateSeparator
+        dateSeparator: @escaping (ChatMessage, ChatMessage) -> Date? = MessageDisplayOptions.defaultDateSeparator,
+        showOriginalTranslatedButton: Bool = false
```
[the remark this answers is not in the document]
nuno-vieira: I can change this yes, no problem 👍

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -7,6 +7,7 @@ import StreamChat
 import SwiftUI
 
 public struct MessageContainerView<Factory: ViewFactory>: View {
+    @ObservedObject var messageViewModel: MessageViewModel
```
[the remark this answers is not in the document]
nuno-vieira: In our case, it does, but in the future, we might want another view to own it (ChannelViewModel), or if a customer wants to use the MessageContainerView directly, and own the view model, it can also be a possibility

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
nuno-vieira: From my texts, all works good. At least regular texts 🤔 Maybe this was related to edge cases?

nuno-vieira: For example this one: https://github.com/GetStream/stream-chat-swiftui/pull/582, works fine

martinmitrevski: I shared one example on slack, let's check that one as well.

nuno-vieira: I've also tested that one, works fine

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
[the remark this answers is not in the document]
laevandus: Especially because we have code like this:
```swift
messageViewModel = viewModel ?? MessageViewModel(
            message: message,
            channel: channel,
            originalTranslationsStore: .shared
        )
```
Meaning it gets created internally and therefore it should be owned by the view.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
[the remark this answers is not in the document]
laevandus: Most of the dependencies are in the Utils class, should it also be there (for consistency sake)?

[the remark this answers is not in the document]
nuno-vieira: You mean moving the store inside the utils? 🤔

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
nuno-vieira: I also thought about this, but doing this now, it will be some significant changes, especially because we need to pass the view model furtherdown at least unil `MessageView`. So for now lets keep it like this to minimize the changes 👍

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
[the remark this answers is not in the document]
nuno-vieira: It complicates things a bit though, because we can't do this, in the initializer, since utils is not available in the initializer:
```
self.originalTranslationsStore.$originalTextMessageIds.sink(
            receiveValue: { [weak self] _ in
                self?.objectWillChange.send()
            }
        )
        .store(in: &cancellables)
```

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageViewModel.swift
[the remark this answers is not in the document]
nuno-vieira: The problem with this, is the LinkDetectioNTextView. We can't do this:
`@ObservedObject var originalTranslationsStore = utils.originalTranslationsStore`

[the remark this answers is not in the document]
nuno-vieira: Ok, there is an alternative, this works:
`@ObservedObject var originalTranslationsStore = InjectedValues[\.utils].originalTranslationsStore`

martinmitrevski (approved):

laevandus (approved):
Looks good

PR #825, opened 2025-05-07
Swift 6: complete concurrency checking

laevandus:
### 🔗 Issue Links

Resolves: [IOS-736](https://linear.app/stream/issue/IOS-736)

### 🎯 Goal

* Set Swift version to 6 and implement complete concurrency checking

### 📝 Summary

* Set Swift version to 6 in StreamChatSwiftUI and demo app (includes `SWIFT_STRICT_CONCURRENCY` = `complete`)
* Add `Sendable` conformance to many types
* Many protocols and types related to UI get `@preconcurrency @MainActor` requirement (e.g. view models)
* Use `StreamConcurrency.onMain` for jumping on the main actor from controller delegates and completion handlers
* Upgrade Nuke to version 12.8 (otherwise we can't compile with complete concurrency checking)
  * LazyImage has breaking changes in Nuke (more testing for image scaling and gifs is needed) 
* Patch SwiftyGif for supporting complete concurrency checking (not beautiful)

### 🧪 Manual Testing Notes

Full manual regression testing round is needed.

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/Appearance.swift
```diff
@@ -21,20 +21,21 @@ public class Appearance {
...
+    public static var localizationProvider: @Sendable(_ key: String, _ table: String) -> String = { key, table in
         Bundle.streamChatUI.localizedString(forKey: key, value: nil, table: table)
     }
 }
 
 // MARK: - Appearance + Default
 
 public extension Appearance {
-    static var `default`: Appearance = .init()
+    nonisolated(unsafe) static var `default`: Appearance = .init()
```
laevandus, line 33: `Appearance` is a class in SwiftUI and its properties are accessible through InjectedValues. 
@Injected makes it impossible top make the `Appearance` itself `@MainActor` because `@Injected(\.images) private var images` can't ensure main actor isolation when a type/view X is created.
The other option would be to use a lock within Appearance to make sure everything is concurrency safe, but that feels like too much because the type itself is called from main anyway (if not mistakenly calling from background thread).

.github/workflows/smoke-checks.yml
```diff
@@ -66,7 +66,8 @@ jobs:
   build-xcode15:
     name: Build SDKs (Xcode 15)
     runs-on: macos-15
-    if: ${{ github.event.inputs.record_snapshots != 'true' }}
+    #if: ${{ github.event.inputs.record_snapshots != 'true' }}
+    if: false
```
laevandus, line 70: Xcode 15 builds are going away, disabling it for now and a cleanup happens separately

Sources/StreamChatSwiftUI/Appearance.swift
```diff
@@ -21,20 +21,21 @@ public class Appearance {
     }
 
     /// Provider for custom localization which is dependent on App Bundle.
-    public static var localizationProvider: (_ key: String, _ table: String) -> String = { key, table in
+    nonisolated(unsafe)
+    public static var localizationProvider: @Sendable(_ key: String, _ table: String) -> String = { key, table in
         Bundle.streamChatUI.localizedString(forKey: key, value: nil, table: table)
     }
```
laevandus, line 27: Alternative could be  making it `@MainActor` instead of opting out with nonisolated

DemoAppSwiftUI/ChannelHeader/ChooseChannelQueryView.swift
```diff
@@ -7,7 +7,9 @@ import StreamChatSwiftUI
 import SwiftUI
 
 struct ChooseChannelQueryView: View {
-    static let queryIdentifiers = ChannelListQueryIdentifier.allCases.sorted(using: KeyPathComparator(\.title))
+    static var queryIdentifiers: [ChannelListQueryIdentifier] {
+        ChannelListQueryIdentifier.allCases.sorted(by: { $0.title < $1.title })
+    }
```
laevandus, line 12: It was giving Sendable related error which did not feel right. Just switched to another sorting method.

DemoAppSwiftUI/LoginViewModel.swift
```diff
@@ -42,7 +42,7 @@ class LoginViewModel: ObservableObject {
                 return
             }
 
-            DispatchQueue.main.async { [weak self] in
+            Task { @MainActor [weak self] in
```
laevandus, line 45: DispatchQueue gives Sendable error for self

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/AddUsersViewModel.swift
```diff
@@ -6,7 +6,7 @@ import StreamChat
 import SwiftUI
 
 /// View model for the `AddUsersView`.
-class AddUsersViewModel: ObservableObject {
+@MainActor class AddUsersViewModel: ObservableObject {
```
laevandus, line 9: Since view model factory is MainActor and view models are used from view, then it made more than sense to make all the view models main actor.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/AddUsersViewModel.swift
```diff
@@ -46,9 +46,11 @@ class AddUsersViewModel: ObservableObject {
         if !loadingNextUsers {
             loadingNextUsers = true
             searchController.loadNextUsers { [weak self] _ in
-                guard let self = self else { return }
-                self.users = self.searchController.userArray
-                self.loadingNextUsers = false
+                MainActor.ensureIsolated { [weak self] in
```
laevandus, line 49: Controllers do not ensure on the compilation level which actor is used. We can use any queue for controller completion handlers although the default is main. Therefore, we need to make sure that main is used. Applies to all the completion handlers and controller delegates.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
```diff
@@ -172,7 +172,7 @@ public class ChatChannelInfoViewModel: ObservableObject, ChatChannelControllerDe
         loadAdditionalUsers()
     }
 
-    public func leaveConversationTapped(completion: @escaping () -> Void) {
+    public func leaveConversationTapped(completion: @escaping @MainActor() -> Void) {
```
laevandus, line 175: For avoiding actor isolation in the view. Better force main on the view model level.

Sources/StreamChatSwiftUI/ChatChannel/Composer/Suggestions/Mentions/MentionsCommandHandler.swift
```diff
@@ -152,15 +152,16 @@ public struct MentionsCommandHandler: CommandHandler {
 
     private func searchAllUsers(for typingMention: String) -> Future<SuggestionInfo, Error> {
         Future { promise in
+            nonisolated(unsafe) let unsafePromise = promise
```
laevandus, line 155: Promise is not sendable even when making SuggestionInfo sendable, added a unsafe check here for bypassing it.

Sources/StreamChatSwiftUI/ChatChannel/Composer/Suggestions/CommandsConfig.swift
```diff
@@ -17,7 +17,7 @@ public protocol CommandsConfig {
     /// Creates the main commands handler.
     /// - Parameter channelController: the controller of the channel.
     /// - Returns: `CommandsHandler`.
-    func makeCommandsHandler(
+    @preconcurrency @MainActor func makeCommandsHandler(
```
laevandus, line 20: Using [backwards compatibility mode](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/incrementaladoption#Backwards-Compatibility) for main actor requirement. In v5 we can clean it up.

Sources/StreamChatSwiftUI/ChatChannel/Composer/ComposerConfig.swift
```diff
@@ -52,7 +52,7 @@ public struct ComposerConfig {
         self.isVoiceRecordingEnabled = isVoiceRecordingEnabled
     }
     
-    public static var defaultAttachmentPayloadConverter: (ChatMessage) -> [AnyAttachmentPayload] = { _ in
+    nonisolated(unsafe) public static var defaultAttachmentPayloadConverter: (ChatMessage) -> [AnyAttachmentPayload] = { _ in
```
laevandus, line 55: Since it is deprecated, I did not put any effort into making it concurrency safe

Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
```diff
@@ -981,7 +995,7 @@ class MessageAttachmentsConverter {
     ///
     /// This operation is synchronous and should only be used if all attachments are already loaded.
     /// Like for example, for draft messages.
-    func attachmentsToAssets(
+    static func attachmentsToAssets(
```
laevandus, line 998: Making it static fixes Sendable issue above where self was required to be Sendable (it is a standalone function)

Sources/StreamChatSwiftUI/ChatChannel/Composer/PhotoAssetsUtils.swift
```diff
@@ -80,8 +80,9 @@ public class PhotoAssetLoader: NSObject, ObservableObject {
 
         exportSession.outputURL = outputURL
         exportSession.outputFileType = .mp4
+        nonisolated(unsafe) let unsafeSession = exportSession
```
laevandus, line 83: There is a Swift 6 friendly export function, but we can't use it because it does not support iOS 13. Suppressing the concurrency error here.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/GiphyAttachmentView.swift
```diff
@@ -103,7 +103,7 @@ struct LazyGiphyView: View {
     var body: some View {
         LazyImage(imageURL: source) { state in
             if let imageContainer = state.imageContainer {
-                NukeImage(imageContainer)
+                Image(uiImage: imageContainer.image)
```
laevandus, line 106: Nuke was upgraded to 12.8 and LazyImage has some breaking changes.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/LinkAttachmentView.swift
```diff
@@ -108,12 +108,18 @@ public struct LinkAttachmentView: View {
...
-                        .priority(.high)
-                        .frame(width: width - 2 * padding, height: (width - 2 * padding) / 2)
-                        .cornerRadius(14)
+                    LazyImage(imageURL: linkAttachment.previewURL ?? linkAttachment.originalURL) { state in
+                        if let image = state.image {
+                            image
+                                .resizable()
+                                .aspectRatio(contentMode: .fill)
+                        }
+                    }
```
laevandus, line 117: LazyImage has [breaking changes](https://github.com/kean/Nuke/blob/main/Documentation/Migrations/Nuke%2012%20Migration%20Guide.md) and requires to be used like above for getting the resize to fill behavior

Sources/StreamChatSwiftUI/ChatChannel/MessageList/ReactionsIconProvider.swift
```diff
@@ -19,7 +18,9 @@ class ReactionsIconProvider {
 
     static func color(for reaction: MessageReactionType, userReactionIDs: Set<MessageReactionType>) -> Color? {
         let containsUserReaction = userReactionIDs.contains(reaction)
-        let color = containsUserReaction ? colors.reactionCurrentUserColor : colors.reactionOtherUserColor
+        let color = containsUserReaction ?
+            InjectedValues[\.colors].reactionCurrentUserColor :
+            InjectedValues[\.colors].reactionOtherUserColor
```
laevandus, line 23: Color is struct and reactionCurrentUserColor is lazy var which triggers mutation. Swift 6 was not happy if colors was static var since then static var needs to be made concurrency safe. Ignored the error by switching to accessing injected values directly.

Sources/StreamChatSwiftUI/ChatChannel/Reactions/MessageActions/DefaultMessageActions.swift
```diff
@@ -20,8 +20,8 @@ public extension MessageAction {
         for message: ChatMessage,
         channel: ChatChannel,
         chatClient: ChatClient,
-        onFinish: @escaping (MessageActionInfo) -> Void,
-        onError: @escaping (Error) -> Void
+        onFinish: @escaping @MainActor(MessageActionInfo) -> Void,
+        onError: @escaping @MainActor(Error) -> Void
```
laevandus, line 24: It is a though one since these functions are used in views, which means that it makes sense to force MainActor. Downside is that it changes public API and ends up changing some view factory methods related to actions (SDK user will get a warning after upgrading the SDK).

Sources/StreamChatSwiftUI/ChatChannel/Reactions/MessageActions/MessageActionsResolver.swift
```diff
@@ -29,20 +29,22 @@ public class MessageActionsResolver: MessageActionsResolving {
         info: MessageActionInfo,
         viewModel: ChatChannelViewModel
     ) {
-        if info.identifier == "inlineReply" {
-            withAnimation {
-                viewModel.quotedMessage = info.message
-                viewModel.editedMessage = nil
+        MainActor.ensureIsolated {
```
laevandus, line 32: TODO: I should probably make the whole type main actor instead.

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelDataSource.swift
```diff
@@ -5,7 +5,7 @@
 import StreamChat
 
 /// Data source providing the chat messages.
-protocol MessagesDataSource: AnyObject {
+@MainActor protocol MessagesDataSource: AnyObject {
```
laevandus, line 8: Since it is used by main actor view model, then makes sense to force it to main actor from the beginning. Note that then delegate implementation does not require nonisolated which makes it cleaner.

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelDataSource.swift
```diff
@@ -53,7 +53,7 @@ protocol ChannelDataSource: AnyObject {
     func loadPreviousMessages(
         before messageId: MessageId?,
         limit: Int,
-        completion: ((Error?) -> Void)?
+        completion: (@Sendable(Error?) -> Void)?
```
laevandus, line 56: TODO: Use main actor because it is called from main actor anyway

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -787,13 +791,15 @@ open class ChatChannelViewModel: ObservableObject, MessagesDataSource {
...
     
     deinit {
-        messageCachingUtils.clearCache()
-        if messageController == nil {
-            utils.channelControllerFactory.clearCurrentController()
-            cleanupAudioPlayer()
-            ImageCache.shared.trim(toCost: utils.messageListConfig.cacheSizeOnChatDismiss)
-            if !channelDataSource.hasLoadedAllNextMessages {
-                channelDataSource.loadFirstPage { _ in }
+        MainActor.ensureIsolated {
```
laevandus, line 794: Isolated deinits are coming, but not available yet

Sources/StreamChatSwiftUI/ChatChannelList/ChannelAvatarsMerger.swift
```diff
@@ -4,31 +4,31 @@
...
 
     /// Placeholder images.
-    private lazy var placeholder1 = images.userAvatarPlaceholder1
-    private lazy var placeholder2 = images.userAvatarPlaceholder2
-    private lazy var placeholder3 = images.userAvatarPlaceholder3
-    private lazy var placeholder4 = images.userAvatarPlaceholder4
+    private var placeholder1: UIImage { images.userAvatarPlaceholder1 }
+    private var placeholder2: UIImage { images.userAvatarPlaceholder2 }
+    private var placeholder3: UIImage { images.userAvatarPlaceholder3 }
+    private var placeholder4: UIImage { images.userAvatarPlaceholder4 }
```
laevandus, line 31: Making everything getter here avoids making lazy var properties concurrency safe. Note that these are indeed called from different threads.

Sources/StreamChatSwiftUI/ChatChannelList/ChannelHeaderLoader.swift
```diff
@@ -36,7 +36,7 @@ open class ChannelHeaderLoader: ObservableObject {
     private var loadedImages = [ChannelId: UIImage]()
     private let didLoadImage = PassthroughSubject<ChannelId, Never>()
 
-    public init() {
+    nonisolated public init() {
```
laevandus, line 39: Otherwise `Utils` can't init it in-place and adding any MainActor isolation to Utils is not feasible.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelHelperViews.swift
```diff
@@ -127,9 +127,9 @@ struct TabBarAccessor: UIViewControllerRepresentable {
 }
 
 var isIphone: Bool {
-    UIDevice.current.userInterfaceIdiom == .phone
+    UITraitCollection.current.userInterfaceIdiom == .phone
```
laevandus, line 130: UIDevice is main actor whereas UITraitCollection is not

Sources/StreamChatSwiftUI/StreamNuke/Nuke/Caching/DataCache.swift
```diff
@@ -45,17 +45,16 @@ final class DataCache: DataCaching, @unchecked Sendable {
     /// The path for the directory managed by the cache.
     let path: URL
 
-    /// The number of seconds between each LRU sweep. 30 by default.
-    /// The first sweep is performed right after the cache is initialized.
-    ///
-    /// Sweeps are performed in a background and can be performed in parallel
-    /// with reading.
-    var sweepInterval: TimeInterval = 30
+    /// The time interval between cache sweeps. The default value is 1 hour.
```
laevandus, line 48: Everything under StreamNuke does not need reviewing since it was Nuke's version upgrade without any manual changes.

Sources/StreamChatSwiftUI/Utils/Errors.swift
```diff
@@ -14,7 +14,7 @@ public struct StreamChatError: Error {
     public let description: String?
 
     /// The additional information dictionary.
-    public let additionalInfo: [String: Any]?
+    public nonisolated(unsafe) let additionalInfo: [String: Any]?
```
laevandus, line 17: It has Any, which can't be Sendable (note that Error is Sendable). Skipping the sendable check here (breaking change to change Any here)

Sources/StreamChatSwiftUI/DefaultViewFactory.swift
```diff
@@ -30,8 +30,8 @@ extension ViewFactory {
     
     public func supportedMoreChannelActions(
         for channel: ChatChannel,
-        onDismiss: @escaping () -> Void,
-        onError: @escaping (Error) -> Void
+        onDismiss: @escaping @MainActor() -> Void,
+        onError: @escaping @MainActor(Error) -> Void
```
laevandus, line 34: This will give a build warning for SDK users who override it in their view factory. Needs to be highlighted in the CHANGELOG

Sources/StreamChatSwiftUI/Utils.swift
```diff
@@ -57,6 +57,7 @@ public class Utils {
         }
     }
 
+    @preconcurrency @MainActor
     public lazy var audioSessionFeedbackGenerator: AudioSessionFeedbackGenerator = StreamAudioSessionFeedbackGenerator()
```
laevandus, line 61: Apple's feedback generator's init is MainActor, therefore it is really difficult to not have our wrapper nonisolated.

StreamChatSwiftUITests/Infrastructure/Mocks/MockBackgroundTaskScheduler.swift
```diff
@@ -7,30 +7,30 @@ import Foundation
 
 /// Mock implementation of `BackgroundTaskScheduler`.
 final class MockBackgroundTaskScheduler: BackgroundTaskScheduler {
-    var isAppActive_called: Bool = false
-    var isAppActive_returns: Bool = true
+    nonisolated(unsafe) var isAppActive_called: Bool = false
```
laevandus, line 10: TODO: review if this can be improved

StreamChatSwiftUITests/Infrastructure/TestTools/ChatClient_Mock.swift
```diff
@@ -22,7 +22,15 @@ public extension ChatClient {
             config: config,
             workerBuilders: [],
             environment: .init(
-                apiClientBuilder: APIClient_Spy.init,
+                apiClientBuilder: {
+                    APIClient_Spy(
```
laevandus, line 26: Sendable warning if not written out with all the arguments

StreamChatSwiftUITests/Tests/StreamChatTestCase.swift
```diff
@@ -21,7 +21,7 @@ open class StreamChatTestCase: XCTestCase {
 
     public var streamChat: StreamChat?
 
-    override open func setUp() {
+    @MainActor override open func setUp() {
```
laevandus, line 24: Many tests want to change main actor isolated types, therefore it is easier to require main actor in a single place

StreamChatSwiftUITestsAppTests/Pages/MessageListPage.swift
```diff
@@ -371,7 +371,7 @@ class MessageListPage {
 
     enum ComposerMentions {
         static var cells: XCUIElementQuery {
-            app.scrollViews["CommandsContainerView"].otherElements.matching(NSPredicate(format: "identifier LIKE 'MessageAvatarView'"))
+            app.scrollViews["CommandsContainerView"].images.matching(NSPredicate(format: "identifier LIKE 'MessageAvatarView'"))
```
laevandus, line 374: Nuke's LazyImage changes causes it

DemoAppSwiftUI/LoginViewModel.swift
martinmitrevski: yes, we should use Task going forward

DemoAppSwiftUI/NotificationsHandler.swift
```diff
@@ -41,25 +41,28 @@ class NotificationsHandler: NSObject, ObservableObject, UNUserNotificationCenter
...
     }
 
     func setupRemoteNotifications() {
         UNUserNotificationCenter
             .current()
-            .requestAuthorization(options: [.alert, .sound, .badge]) { granted, _ in
+            .requestAuthorization(options: [.alert, .sound, .badge]) { @Sendable granted, _ in
                 if granted {
-                    DispatchQueue.main.async {
+                    Task { @MainActor in
```
martinmitrevski, line 65: good to test this part, I remember having crashes on video in the summer (probably it was Xcode bug, but still)

Sources/StreamChatSwiftUI/Appearance.swift
martinmitrevski: this is fine

Sources/StreamChatSwiftUI/Appearance.swift
martinmitrevski: also fine to keep it like this, it anyway contains only UI stuff - shouldn't be called from a bg thread

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
```diff
@@ -7,7 +7,7 @@ import StreamChat
 import SwiftUI
 
 // View model for the `ChatChannelInfoView`.
-public class ChatChannelInfoViewModel: ObservableObject, ChatChannelControllerDelegate {
+@preconcurrency @MainActor public class ChatChannelInfoViewModel: ObservableObject, ChatChannelControllerDelegate {
```
martinmitrevski, line 10: why @preconcurrency?

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
```diff
@@ -249,19 +255,21 @@ public class ChatChannelInfoViewModel: ObservableObject, ChatChannelControllerDe
...
-                    ParticipantInfo(
-                        chatUser: member,
-                        displayName: member.name ?? member.id,
-                        onlineInfoText: self.onlineInfo(for: member),
-                        isDeactivated: member.isDeactivated
-                    )
-                }
-                if newMembers.count > self.participants.count {
-                    self.participants = newMembers
+            MainActor.ensureIsolated { [weak self] in
```
martinmitrevski, line 258: this crashes if it's not on the main actor, right? Do we really need that?

Sources/StreamChatSwiftUI/ChatChannel/Composer/Suggestions/Mentions/MentionsCommandHandler.swift
martinmitrevski: we should probably refactor this to use async

Sources/StreamChatSwiftUI/ChatChannel/MessageList/GiphyAttachmentView.swift
```diff
@@ -103,7 +103,12 @@ struct LazyGiphyView: View {
     var body: some View {
         LazyImage(imageURL: source) { state in
             if let imageContainer = state.imageContainer {
-                NukeImage(imageContainer)
+                if imageContainer.type == .gif {
```
martinmitrevski, line 106: why do we change this?

Sources/StreamChatSwiftUI/ChatChannel/MessageList/GiphyAttachmentView.swift
```diff
@@ -119,3 +124,18 @@ struct LazyGiphyView: View {
         .aspectRatio(contentMode: .fit)
     }
 }
+
+private struct AnimatedGifView: UIViewRepresentable {
```
martinmitrevski, line 128: did you take this from Nuke or is it a new implementation?

Sources/StreamChatSwiftUI/DefaultViewFactory.swift
martinmitrevski: ouch, that's tricky, can we avoid it?

DemoAppSwiftUI/ChannelHeader/NewChatViewModel.swift
```diff
@@ -99,31 +99,37 @@ class NewChatViewModel: ObservableObject, ChatUserSearchControllerDelegate {
...
+                    self.loadingNextUsers = false
+                }
             }
         }
     }
 
     // MARK: - ChatUserSearchControllerDelegate
 
-    func controller(
+    nonisolated func controller(
```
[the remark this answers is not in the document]
laevandus: `nonisolated` is required because I made the view model `@MainActor` which I feel like makes sense to have. Controllers have a design where I can set any queue to be the one used for delegate callbacks. In 99% of cases it is the default main queue (highly unlikely that many set their own queues). But because we support any queue for delegate callbacks, we can't make the delegate protocol `@MainActor`. We can't use `@preconcurrency` for the delegate conformance because then we would get runtime crash if anyone change the the default queue in the controller to be a background queue.
Ideally we would use our async-await state layer instead of delegates in the future, but that is out of scope.

Summary is that nonisolated can be avoided if view models are not `@MainActor` (I currenty changed them to be). That said, in the UIKit implementation view controllers are implementing delegates, so nonisolated is required there because UIViewController is `@MainActor` (similar situation). Similar issue with completion handlers where the calling thread can be any thread.

DemoAppSwiftUI/AppleMessageComposerView.swift
```diff
@@ -214,7 +214,7 @@ struct BlurredBackground: View {
 }
 
 struct HeightPreferenceKey: PreferenceKey {
-    static var defaultValue: CGFloat? = nil
+    static let defaultValue: CGFloat? = nil
```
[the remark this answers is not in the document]
laevandus: `PreferenceKey`'s protocol requires the default value to be `get`, therefore this is fine to be defined with let over mutable var (which causes concurrency errors).
[Docs](https://developer.apple.com/documentation/swiftui/preferencekey/defaultvalue)

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
laevandus: I decided to use `@preconcurrency @MainActor` for public types for avoiding build errors for SDK users who are not using complete concurrency checking in their module. In v5 we would remove it. Otherwise their app would fail to build if they use minimal checking and interact/create this type.
[Docs](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/incrementaladoption/#Backwards-Compatibility)

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
laevandus: It is a custom addition (maybe we should rename it to be more clear about it) which makes sure the closure runs on the main thread. Note that one can set any queue to be the one used for completion handers (default is main). This custom addition just checks if we are currently on the main, if yes, just call the closure, otherwise jump on the main thread (since UI updates and MainActor view models require main).

The difference between:
`Task { @MainActor in` and `MainActor.ensureIsolated` is that the latter is quicker.

```swift
DispatchQueue.global().async {
            print("Simulate API calls and DB updates")
            DispatchQueue.main.async {
                print(Thread.isMainThread)
                Task { @MainActor in
                    print("Task main actor")
                }
                MainActor.ensureIsolated {
                    print("Ensure isolated")
                }
            }
        }
```
If this snippet runs, ensure isolated is called first.

I did some micro benchmarking just for fun and the difference is insignificant: we are talking about 1-10 microseconds. Therefore, I guess it makes sense to simplify this and get rid of the custom ensureIsolated.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/GiphyAttachmentView.swift
laevandus: Nuke dropped their image type and gif rendering support in Nuke 12.0 ([migration guide](https://github.com/kean/Nuke/blob/main/Documentation/Migrations/Nuke%2012%20Migration%20Guide.md#lazyimage)). This is their recommendation how to handle gif rendering.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/GiphyAttachmentView.swift
laevandus: This I took from SwiftyGif's [documentation](https://github.com/alexiscreuzot/SwiftyGif?tab=readme-ov-file#swiftui) as a recommended way for rendering gifs in SwiftUI.

Sources/StreamChatSwiftUI/DefaultViewFactory.swift
laevandus: I'll take it as a TODO to find a way to avoid it.

laevandus: I went for reverting these changes since these will cause breaking changes for customers. Required some nonisolated(unsafe) handling.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
laevandus: I renamed the custom addition to StreamConcurrency.onMain. I tried to get rid of it, but loads of unit-tests will break because using a Task introduces slight delay.

DemoAppSwiftUI/NotificationsHandler.swift
laevandus: I got the same crash and the recommendation was to make it explicitly @Sendable.

Sources/StreamChatSwiftUI/ChatChannel/Composer/Suggestions/Mentions/MentionsCommandHandler.swift
laevandus: Can happen in v5, currently the Future type gets exposed with public function:
```swift
    public func showSuggestions(
        for command: ComposerCommand
    ) -> Future<SuggestionInfo, Error> {
        showMentionSuggestions(
            for: command.typingSuggestion.text,
            mentionRange: command.typingSuggestion.locationRange
        )
    }
```

Sources/StreamChatSwiftUI/ChatChannel/Reactions/MessageActions/DefaultMessageActions.swift
```diff
@@ -299,12 +299,14 @@ public extension MessageAction {
             messageId: message.id
         )
 
+        nonisolated(unsafe) let unsafeOnError = onError
+        nonisolated(unsafe) let unsafeOnFinish = onFinish
```
laevandus, line 303: Opting out for avoiding breaking changes in the public API (can be fixed properly in v5)

Sources/StreamChatSwiftUI/ChatChannel/Composer/Suggestions/Mentions/MentionsCommandHandler.swift
martinmitrevski: maybe we can use async/await under the hood, and just wrap the result in a Future?

martinmitrevski: anyway, low prio

Sources/StreamChatSwiftUI/Utils/StreamConcurrency.swift
```diff
@@ -0,0 +1,28 @@
...
+enum StreamConcurrency {
+    /// Synchronously performs the provided action on the main thread.
+    ///
+    /// Used for ensuring that we are on the main thread when compiler can't know it. For example,
+    /// controller completion handlers by default are called from main thread, but one can
+    /// configure controller to use background thread for completions instead.
+    ///
+    /// - Important: It is safe to call from any thread. It does not deadlock if we are already on the main thread.
+    /// - Important: Prefer Task { @MainActor if possible.
+    static func onMain<T>(_ action: @MainActor @Sendable() throws -> T) rethrows -> T where T: Sendable {
```
martinmitrevski, line 16: why don't we make the one from LLC public and use only that one?

martinmitrevski (commented):
Third-party dependencies have the most changes here - we should thoroughly test it. Performance impact is also something we should check.

laevandus:
I am going to close this PR for now since it will have major changes based on what we have discussed.

PR #833, opened 2025-05-21
Show dynamically time, relative date, weekday, or short date in channel lists

laevandus:
### 🔗 Issue Links

Resolves [IOS-851](https://linear.app/stream/issue/IOS-851)

### 🎯 Goal

* Show time, relative date, weekday or short date in channel lists instead of just time of the message (disabled by default, can be enabled with `utils.channelListConfig.messageRelativeDateFormatEnabled = true`)

### 📝 Summary

* Add `ChannelListDateFormatter` for dynamic date formatting (based on the [UIKit implementation](https://github.com/GetStream/stream-chat-swift/pull/2736))
* Add `Utils.channelListDateFormatter` for overriding the formatting behaviour
* Add `ChannelListConfig` for channel specific settings
* Use the new formatter in channel list and channel search (if `ChannelListConfig.messageRelativeDateFormatEnabled` is set to true)
* Enabled the behaviour in the demo app

### 🛠 Implementation

Logic:
Today: only time, e.g 12:10
Yesterday: "Yesterday"
Last 7 days: Weekday e.g. "Monday"
Older: short date, e.g. 21.05.2025

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  BeforeChannels   |  AfterChannels  |
|  BeforeSearch   |  AfterSearch  |

### 🧪 Manual Testing Notes

Open channel list and observe last message dates
Search for channels and observe last message dates

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageContainerView.swift
```diff
@@ -36,6 +36,7 @@ public struct MessageContainerView<Factory: ViewFactory>: View {
     private var paddingValue: CGFloat {
         utils.messageListConfig.messagePaddings.singleBottom
     }
+
```
laevandus, line 39: Missed linter issue when I merged one of the external contributions

Sources/StreamChatSwiftUI/Utils/Common/ChannelListDateFormatter.swift
```diff
@@ -0,0 +1,69 @@
+//
+// Copyright © 2025 Stream.io Inc. All rights reserved.
+//
+
+import Foundation
+
+/// A formatter that converts last message timestamps in the channel list.
+///
+/// Shows time, relative date, weekday or short date based on days passed.
+public final class ChannelListDateFormatter: DateFormatter, @unchecked Sendable {
```
laevandus, line 10: All of this is based on UIKit's implementation

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### ✅ Added
 - Add extra data to user display info [#819](https://github.com/GetStream/stream-chat-swiftui/pull/819)
 - Make message spacing in message list configurable [#830](https://github.com/GetStream/stream-chat-swiftui/pull/830)
+- Show time, relative date, weekday, or short date for last message in channel list and search [#833](https://github.com/GetStream/stream-chat-swiftui/pull/833)
```
nuno-vieira, line 9: This should be move to Changed section, since it is a change in behaviour, and customers should be warned about it. (The ones who use the default formatter)

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
```diff
@@ -337,7 +337,7 @@ extension ChatChannel {
 
     public var timestampText: String {
         if let lastMessageAt = lastMessageAt {
-            return InjectedValues[\.utils].dateFormatter.string(from: lastMessageAt)
+            return InjectedValues[\.utils].channelListDateFormatter.string(from: lastMessageAt)
```
nuno-vieira, line 340: We can't really do this, otherwise the customers that are providing a customer date formatter already, will have a breaking change in the UI. 

The ticket description suggests that we should subclass the existing DateFormatter and provide the new logic. Did you try this? It should be a question of creating a subclass of DateFormatter, and internally use the UIKit implementation 🤔

nuno-vieira (changes requested):
I'm requesting changes since this would be a breaking UI change for customers, we need to try to use the same `utils.dateFormatter`

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
laevandus: `Utils.dateFormatter` is used in many places, but the new logic should only be for channel list (and channel search). Subclassing the existing formatter would create the same breaking change situation and would be harder to manage due to the fact that sometimes we need to keep the old behaviour, sometimes the new logic. I'll look into adding a flag for this.

nuno-vieira: I see.. kinda tricky then... lets discuss this internally

CHANGELOG.md
laevandus: Added a flag which makes it disabled by default.

nuno-vieira (approved):
LGTM! ✅

martinmitrevski (approved):
looks good ✅  probably better to have it on by default, but leaving it up to you two to decide.

PR #839, opened 2025-05-27
Allow customisation of message previews in lists

laevandus:
### 🔗 Issue Links

Resolves: [IOS-895](https://linear.app/stream/issue/IOS-895)

### 🎯 Goal

* Allow overriding message previews in lists (e.g. supporting previews for custom attachments)

### 📝 Summary

* Open up the `MessagePreviewFormatter`

### 🛠 Implementation

* Open up existing `MessagePreviewFormatter`

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/Utils/MessagePreviewFormatter.swift
```diff
@@ -5,23 +5,33 @@
...
 
+/// Provides message preview representation for lists.
+public protocol MessagePreviewFormatting {
+    /// Formats the message including the author's name.
+    func format(_ previewMessage: ChatMessage, in channel: ChatChannel) -> String
+    /// Formats only the content of the message without the author's name.
+    func formatContent(for previewMessage: ChatMessage, in channel: ChatChannel) -> String
+    /// Formats only the attachment content of the message in case it contains attachments.
+    func formatAttachmentContent(for previewMessage: ChatMessage, in channel: ChatChannel) -> String?
+}
```
nuno-vieira, line 16: I think it is probably better to just use the `MessagePreviewFormatter` function directly here 🤔 There is a chance we add new function here, for example, for Draft Message etc... and if we introduce a new function, it will be a breaking change.

nuno-vieira: The other formatters, we use a protocol, but usually it is just one function, and for one use case. This one is a bit different, since we use it in multiple places, and there is a chance to extend it further

martinmitrevski: I it's also fine like this, you can add a protocol extension with a default implementation to avoid breaking changes.

martinmitrevski (approved):

Sources/StreamChatSwiftUI/Utils/MessagePreviewFormatter.swift
laevandus: I did consider skipping the protocol part completely, but then only added it because many existing types in Utils had it. I guess it makes sense.

nuno-vieira: Yeah the default extension is a good point, forgot about it. In that case I'm fine with both approaches 👍 maybe let's stick with the protocol for consistency

PR #870, opened 2025-07-01
Fix channel view tab bar not hidden on iOS 16.0

nuno-vieira:
### 🔗 Issue Links
https://linear.app/stream/issue/IOS-966/fix-channel-view-tab-bar-not-hidden-in-ios-160

### 🎯 Goal

Fix channel view tab bar not hidden on iOS 16.0.

### 📝 Summary

- Fix channel view tab bar not hidden on iOS 16.0
- Make sure the Thread View also uses the `messageListConfig.handleTabBarVisibility`

### 🛠 Implementation

Add the Thread List tab bar visibility logic to the Channel List as well.

Since the Demo App uses a custom navigation item view, I used an existing, simpler modifier.

### 🧪 Manual Testing Notes

1. Install iOS 16.0 Simulator on Xcode
2. Run the app on iOS 16.0
3. Open a channel
4. The tab bar should be visible.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
@@ -14,6 +14,7 @@ public struct ChatChannelNavigatableListItem<Factory: ViewFactory, ChannelDestin
     private var avatar: UIImage
     private var disabled: Bool
     private var onlineIndicatorShown: Bool
+    private var handleTabBarVisibility: Bool
     @Binding private var selectedChannel: ChannelSelectionInfo?
     private var channelDestination: (ChannelSelectionInfo) -> ChannelDestination
     private var onItemTap: (ChatChannel) -> Void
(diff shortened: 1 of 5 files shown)


Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
```diff
@@ -57,7 +60,10 @@ public struct ChatChannelNavigatableListItem<Factory: ViewFactory, ChannelDestin
                 tag: channel.channelSelectionInfo,
                 selection: $selectedChannel
             ) {
-                LazyView(channelDestination(channel.channelSelectionInfo))
+                LazyView(
+                    channelDestination(channel.channelSelectionInfo)
+                        .modifier(HideTabBarModifier(handleTabBarVisibility: handleTabBarVisibility))
```
martinmitrevski, line 65: We already have handling for the tab bar, somewhere, this would duplicate the handling. Maybe we just optionally add it between 16.0 and 16.3?

nuno-vieira: This modifier is being used on Thread List, I'm reusing this one. The other one that we have is simpler and does not work on every scenario, and it is on the Demo App only

martinmitrevski: You need to test this with <iOS 16 simulator. If that works, I'm fine with removing all that code.

martinmitrevski (approved):

PR #956, opened 2025-09-15
Fix composer not showing images in the composer when editing signed attachments

nuno-vieira:
### 🔗 Issue Links

IOS-1014
IOS-817

### 🎯 Goal

- Fix composer not showing images in the composer when editing signed attachments
- Fix replacing an image while editing a message not showing the new image in the message list
- Removes dead code

### 🧪 Manual Testing Notes

**Fix composer not showing images in the composer when editing signed attachments:**
N/A

**Fix replacing an image while editing a message not showing the new image in the message list:**
1. Send a message with an image
2. Long tap the message
3. Edit the message
4. Remove the existing image and upload a new one
5. Send the message
6. The new message should appear while uploading and also when finished

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### 🐞 Fixed
 - Fix updating back button tint with `ColorPalette.navigationBarTintColor` [#953](https://github.com/GetStream/stream-chat-swiftui/pull/953)
 - Fix swipe to reply enabled when quoting a message is disabled  [#977](https://github.com/GetStream/stream-chat-swiftui/pull/957)
+- Fix composer not showing images in the composer when editing signed attachments [#956](https://github.com/GetStream/stream-chat-swiftui/pull/956)
+- Fix replacing an image while editing a message not showing the new image in the message list [#956](https://github.com/GetStream/stream-chat-swiftui/pull/956)
 
 # [4.88.0](https://github.com/GetStream/stream-chat-swiftui/releases/tag/4.88.0)
 _September 10, 2025_


Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
```diff
@@ -971,59 +971,156 @@ class MessageAttachmentsConverter {
...
+        _ attachment: AnyChatMessageAttachment,
+        completion: @escaping (AddedAsset?) -> Void
+    ) {
+        guard let imageAttachment = attachment.attachment(payloadType: ImageAttachmentPayload.self) else {
+            return completion(nil)
+        }
+
+        // If there is a custom CDN, we need to load the image from the CDN
+        // especially if the images are signed.
+        if !(utils.imageCDN is StreamImageCDN) {
```
martinmitrevski, line 1077: this is a bit unsafe if we decide to provide a new implementation

nuno-vieira:
> few issues:
> 
> * seems like both the keyboard and the image attachments are shown at the same time, leading to this UI:
> 
> IMG_4343 - if I start editing, leave the screen, it treats it as a draft which if I send it - it sends it as a new message. Probably, we should disable saving a draft when it's an edited message?

Yeah, that UI is a known issue that is in the backlog. Discarding the draft when editing is questionable 🤔 On UIKit at the moment we have this behaviour. In one way, if you are editing a long message and accidentally leave the channel, it is kinda nice that we save the edit. At the moment, the reference of the edited message is not saved, so the user would need to tap on edit again. So yeah at the moment even if we wanted to save the draft edited message, it is incomplete 🤔 

Update:
For now, I removed the draft when editing a message

Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
nuno-vieira: Fixed ✅

PR #973, opened 2025-09-23
[CI] Update swift format lane to check for foundation import

testableapple:
### 🎯 Goal

Check for Foundation import where required on swiftformat action

### 📝 Summary

The CI will alert if there is no Foundation import (UIKit and StreamChat imports are an exception in this case)

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -491,9 +491,25 @@ lane :lint_pr do
   danger(dangerfile: 'Dangerfile') if is_ci
 end
 
+def check_foundation_import
+  files_missing_imports = []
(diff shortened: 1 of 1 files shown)


Sources/StreamChatSwiftUI/StreamSwiftyGif/NSImage+SwiftyGif.swift
```diff
@@ -4,6 +4,7 @@
 
 #if os(macOS)
 
+import Foundation
```
martinmitrevski, line 7: We also don't need it in third party dependencies

PR #1013, opened 2025-10-14
[Feature] Add the `makeAttachmentTextView` method to ViewFactory

sheep-q:
### 🔗 Issue Links

N/A

### 🎯 Goal

Be able to customize the text view for messages that contain both attachments and text.

### 📝 Summary

Add a method to the ViewFactory that makes the attachment text view 

### 🛠 Implementation

Add the `makeAttachmentTextView` method to ViewFactory to enable customization

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/ImageAttachmentView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/ImageAttachmentView.swift
@@ -101,7 +101,7 @@ public struct AttachmentTextView<Factory: ViewFactory>: View {
     var message: ChatMessage
     let injectedBackgroundColor: UIColor?
 
-    public init(factory: Factory, message: ChatMessage, injectedBackgroundColor: UIColor? = nil) {
+    public init(factory: Factory = DefaultViewFactory.shared, message: ChatMessage, injectedBackgroundColor: UIColor? = nil) {
         self.factory = factory
         self.message = message
         self.injectedBackgroundColor = injectedBackgroundColor
--- a/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
@@ -18,7 +18,7 @@ public struct MessageView<Factory: ViewFactory>: View {
     public var isFirst: Bool
     @Binding public var scrolledId: String?
 
-    public init(factory: Factory = DefaultViewFactory.shared, message: ChatMessage, contentWidth: CGFloat, isFirst: Bool, scrolledId: Binding<String?>) {
+    public init(factory: Factory, message: ChatMessage, contentWidth: CGFloat, isFirst: Bool, scrolledId: Binding<String?>) {
         self.factory = factory
         self.message = message
         self.contentWidth = contentWidth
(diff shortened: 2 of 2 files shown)


Sources/StreamChatSwiftUI/ChatChannel/MessageList/ImageAttachmentView.swift
```diff
@@ -93,21 +93,23 @@ public struct ImageAttachmentContainer<Factory: ViewFactory>: View {
...
+public struct AttachmentTextView<Factory: ViewFactory>: View {
     @Injected(\.colors) private var colors
     @Injected(\.fonts) private var fonts
 
+    var factory: Factory
     var message: ChatMessage
     let injectedBackgroundColor: UIColor?
 
-    public init(message: ChatMessage, injectedBackgroundColor: UIColor? = nil) {
+    public init(factory: Factory, message: ChatMessage, injectedBackgroundColor: UIColor? = nil) {
```
martinmitrevski: We should put a default parameter `DefaultViewFactory.shared` to avoid breaking changes here.

Sources/StreamChatSwiftUI/DefaultViewFactory.swift
```diff
@@ -1151,6 +1151,12 @@ extension ViewFactory {
     ) -> some View {
         AddUsersView(loadedUserIds: options.loadedUsers.map(\.id), onUserTap: onUserTap)
     }
+    
+    public func makeAttachmentTextView(
+        message: ChatMessage
```
martinmitrevski: We now follow an approach where we put an options class instead of the params directly. So this would be something like `AttachmentTextViewOptions`, with one property `message` (to enable adding new properties in the future).

martinmitrevski (commented):
Looks good, left few small comments.

sheep-q:
Thanks @martinmitrevski for the review. Pls take another look at new updates

Sources/StreamChatSwiftUI/DefaultViewFactory.swift
martinmitrevski: no, just public is enough - we would like to control the extensibility of these classes internally.

Sources/StreamChatSwiftUI/ChatChannel/MessageList/ImageAttachmentView.swift
martinmitrevski: we should put the default param here

Sources/StreamChatSwiftUI/ChatChannel/MessageList/MessageView.swift
```diff
@@ -247,6 +247,16 @@ struct StreamTextView: View {
     }
 }
 
+// Options for the attachment text view.
+open class AttachmentTextViewOptions {
```
martinmitrevski: this should be public

sheep-q: Sure. Pls take another look. Thank you

martinmitrevski (approved):
Thanks for the contribution! Looks good from my side ✅ @testableapple can you please run the checks before merging?

PR #1050, opened 2025-11-13
Fix Throttler crash in `ChatChannelViewModel.handleMessageAppear()`

nuno-vieira:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1245

### 🎯 Goal

Fix Throttler crash in `ChatChannelViewModel.handleMessageAppear()`.

### 📝 Summary
- Fix Throttler crash in `ChatChannelViewModel.handleMessageAppear()`
- Remove unnecessary channel query call when leaving the channel view in a mid-page

#### Problem:
The crash was caused by accessing the channel controller at the same time by the main thread and the Throttler's serial queue.

#### Solution:
We need to change the Throttler's queue to be the main thread as well, just like we are doing in UIKit.

### 🧪 Manual Testing Notes

N/A. Unit tests have been added to replicate the crash.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Expose `FilePickerView.init(fileURLs:)` [#1049](https://github.com/GetStream/stream-chat-swiftui/pull/1049)
 - Add `MessageComposerViewModel.updateAddedAssets()` [#1049](https://github.com/GetStream/stream-chat-swiftui/pull/1049)
 
+### 🐞 Fixed
+- Fix `Throttler` crash in `ChatChannelViewModel.handleMessageAppear()` [#1050](https://github.com/GetStream/stream-chat-swiftui/pull/1050)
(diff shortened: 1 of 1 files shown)


StreamChatSwiftUITests/Tests/ChatChannel/ChatChannelViewModel_Tests.swift
```diff
@@ -772,6 +772,31 @@ class ChatChannelViewModel_Tests: StreamChatTestCase {
...
+                viewModel.handleMessageAppear(index: 0, scrollDirection: .down)
+                viewModel.handleMessageAppear(index: 0, scrollDirection: .down)
+                viewModel.handleMessageAppear(index: 0, scrollDirection: .down)
+                viewModel.handleMessageAppear(index: 0, scrollDirection: .down)
+                viewModel.handleMessageAppear(index: 0, scrollDirection: .down)
+            }
+        }
+        
+        // Then - Should not crash
+        XCTAssert(true)
```
laevandus: Seems like no reason to add, if crashes, then the test will crash and fail.

PR #1059, opened 2025-11-27
Fix channel list skipping some updates on iPad

laevandus:
### 🔗 Issue Links

Fixes: [IOS-1246](https://linear.app/stream/issue/IOS-1246)

### 🎯 Goal

Fix channel list skipping some updates on iPad

### 📝 Summary

* Remove queued updates array which can be stale when channels are updating often

### 🛠 Implementation

Channel list view model has optimised channel list updates implemented. The logic is that when channel is selected or search is displayed, updates are not applied to the channel list, when selection is reset, then queued channels are applied (simplified description). Problem is that the queued channels array can get out of sync due to if logic in handle channels changed. Moreover, another downside is that split view should make sure that `updatedChannelsFromMessageList` is true, otherwise updates are skipped although channel list is visible while channel is selected. It felt a bit too complex and I tried to simplify this by always reading channels from the controller and using just a bool for optimised state. Also, not `updatedChannelsFromMessageList` does not need to be set to true on iPad because the logic checks if split view config is true or false.

### 🧪 Manual Testing Notes

Run demo app on iPad:
1. Select a channel on A
2. Use another simulator to trigger channel list updates for A
3. A sees that channels update

Run demo app on iPad again, but set iPad split view config to false and updateChannelsFromMessageList to false as well which changes the navigation view to use stacked style like on iPhone.
```
let config = MessageListConfig(updateChannelsFromMessageList: false, iPadSplitViewEnabled: false)
        streamChat = StreamChat(chatClient: chatClient, utils: Utils(messageListConfig: config))
```
1. Select a channel on A
2. Use another simulator to trigger channel list updates for A
3. A sees that channels are up to date after navigating back to the channel list

Note: unit test was added for optimised updates, no need to verify that manually.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Expose `QuotedMessageViewContainer` [#1056](https://github.com/GetStream/stream-chat-swiftui/pull/1056)
 - Add `QuotedMessageContentView` and `ViewFactory.makeQuotedMessageContentView()` [#1056](https://github.com/GetStream/stream-chat-swiftui/pull/1056)
 - Allow customizing the attachment size and avatar size of the quoted message view [#1056](https://github.com/GetStream/stream-chat-swiftui/pull/1056)
+### 🐞 Fixed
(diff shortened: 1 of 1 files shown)


laevandus:
Got a confirmation that it is OK for them as well.

PR #1200, opened 2026-02-13
v5: [Composer Redesign] Images, Videos and Files Attachments

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1368/v5-composer-redesign-image-attachment-preview
https://linear.app/stream/issue/IOS-1372/v5-composer-redesign-documents-attachment-preview
https://linear.app/stream/issue/IOS-1371/v5-composer-redesign-mixed-attachments-preview-refactor
https://linear.app/stream/issue/IOS-1369/v5-composer-redesign-video-attachment-preview

### 🎯 Goal

Redesign the image, video and file attachments in the composer and render them as part of the same container.

### 🛠 Implementation

- New `ComposerImageAttachmentView`
- New `ComposerFileAttachmentView`
- New `ComposerVideoAttachmentView`
- New `MediaBadge`
- New unified `ComposerAttachmentsContainerView` that renders all attachments in one horizontal row
- New `ComposerAsset` model  to support rendering all attachment types in one container
- Renamed the previous `ComposerAssets` to `TotalAddedAssets` to not confuse with the previous one
- Moved Message List and Composer source files to top-level folders

**Note:**
Overall the the `AddedAsset`, `AddedVoiceRecordings`, `AddedFileURLs` etc... it is quite confusing at the moment. I tried to do a refactoring but it was too many changes, so for now it is out of scope. Ideally the composer should follow the same approach of the UIKit SDK, and use directly the existing attachment types from LLC, since they support type eraser out of the box, and it will make it easier to support the unified `ComposerAttachmentsContainerView` without creating an enum.

**Out of scope:**
- Voice Recordings
- AddedAsset Refactoring

### 🎨 Showcase

https://github.com/user-attachments/assets/20afbf21-2634-426b-9266-f72aa7ff746d

### 🧪 Manual Testing Notes

Smoke testing of Images, Videos and Files in the composer.

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerFileAttachmentView.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private var previewImage: UIImage {
+        let iconName = url.pathExtension
+        return images.fileIconPreviews[iconName] ?? images.fileFallback
+    }
+}
+
+private extension URL {
+    var sizeString: String {
```
martinmitrevski, line 65: we should check if this can impact performance, I imagine it would be called many times

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerImageAttachmentView.swift
```diff
@@ -0,0 +1,31 @@
...
+            .resizable()
+            .aspectRatio(contentMode: .fill)
+            .frame(width: imageSize, height: imageSize)
+            .clipped()
+            .cornerRadius(tokens.messageBubbleRadiusAttachment)
+            .overlay(
+                RoundedRectangle(cornerRadius: tokens.messageBubbleRadiusAttachment)
+                    .strokeBorder(Color(colors.borderCoreOpacity10), lineWidth: 1)
+            )
+            .id(attachment.id)
```
martinmitrevski, line 26: why do we need the id?

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
```diff
@@ -0,0 +1,40 @@
...
+            attachment: attachment,
+            onDiscardAttachment: onDiscardAttachment
+        )
+        .mediaBadgeOverlay {
+            VideoMediaBadge(durationText: videoDurationText)
+        }
+    }
+
+    private var videoDurationText: String {
+        guard let raw = attachment.extraData["duration"] else { return "0s" }
```
martinmitrevski: shouldn't we localize these?

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
```diff
@@ -0,0 +1,40 @@
...
+        guard case let .string(durationString) = raw else { return "0s" }
+        let seconds = parseDurationString(durationString)
+        return "\(seconds)s"
+    }
+
+    private func parseDurationString(_ string: String) -> Int {
+        let parts = string.split(separator: ":")
```
martinmitrevski: also, maybe we can consider putting these out of the view, so we can write some tests?

Sources/StreamChatSwiftUI/ChatComposer/Attachments/MediaBadge.swift
```diff
@@ -0,0 +1,149 @@
...
+        self.icon = icon
+        self.durationText = durationText
+    }
+
+    public var body: some View {
+        HStack(spacing: tokens.spacingXxs) {
+            icon
+                .frame(height: tokens.iconSizeXs)
+            Text(durationText)
+                .font(.system(size: 10, weight: .bold))
```
martinmitrevski, line 67: why fixed font?

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerView.swift
```diff
@@ -152,7 +150,7 @@ public struct MessageComposerView<Factory: ViewFactory>: View, KeyboardReadable
                     attachmentPickerState: $viewModel.pickerState,
                     filePickerShown: $viewModel.filePickerShown,
                     cameraPickerShown: $viewModel.cameraPickerShown,
-                    addedFileURLs: $viewModel.addedFileURLs,
+                    addedFileURLs: viewModel.addedFileURLsBinding,
```
martinmitrevski: why Binding suffix?

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
```diff
@@ -13,12 +13,44 @@ import SwiftUI
...
+    @Published public var composerAssets = [ComposerAsset]() {
+        didSet {
+            if shouldDeleteDraftMessage(oldValue: oldValue) {
+                deleteDraftMessage()
+            }
+        }
+    }
+
+    /// Binding for the file picker — appends selected files to composerAssets.
+    public var addedFileURLsBinding: Binding<[URL]> {
```
martinmitrevski: yes, the naming is not ideal - binding is implementation detail, e.g. could be observable in the future. 
Also, I think the solution itself is overly complex:
- we duplicate the logic for getting the composer assets (I also saw this being used elsewhere, so good to abstract it)
- also, maybe a better way would've been to make this computed variable and expose a setter?

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
```diff
@@ -483,44 +493,27 @@ import SwiftUI
...
+                }
+            case .addedFile:
+                newAssets.append(composerAsset)
             }
         }
-        
-        if !imageRemoved && canAddAttachment(with: addedAsset.url) {
-            images.append(addedAsset)
+        if !found && canAddAttachment(with: addedAsset.url) {
+            newAssets.insert(.addedAsset(addedAsset), at: 0)
```
martinmitrevski: how did the animation look like when appended at the end?

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
```diff
@@ -651,19 +639,38 @@ import SwiftUI
...
+                    return FileAddedAsset(url: url, payload: addedRemoteFileURLs[url])
+                }
+                return nil
+            },
+            voiceAssets: addedVoiceRecordings,
+            customAssets: addedCustomAttachments
+        )
+    }
+
+    private func updateComposerAssets(_ assets: TotalAddedAssets) {
```
martinmitrevski, line 643: for example, the approach I suggested above is used here (although the method here is private) - good if we have consistency

martinmitrevski (commented):
Additionally, there are few other issues:
- why are some snapshots from the composer bigger now? e.g. StreamChatSwiftUITests/Tests/ChatChannel/__Snapshots__/MessageComposerView_Tests/test_composerInputView_command.default-light-themed.png
- there are many conflicts with the base branch, something feels off here

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerFileAttachmentView.swift
nuno-vieira: This was already there, I did not change anything here, just moved from a different file, so if it was working before, I'm assuming it should still work

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerImageAttachmentView.swift
nuno-vieira: It was used in the old view, so I also used it here, not sure why it was there, so I kept the same thing

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
nuno-vieira: This is more about formatting, not localisation, we can clean this later when we have a better picture of the formatters. This then should go to the common module

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
nuno-vieira: Same as above, the goal is to move this to formatters, I can do this as part of this PR if you want, but the figma is not yet consistent on how to format video durations

Sources/StreamChatSwiftUI/ChatComposer/Attachments/MediaBadge.swift
nuno-vieira: It should be `iconSizeXs` actually, but it should be fixed here, we don't want the badge to increase in this case since the size is fixed as well

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerView.swift
nuno-vieira: AI renaming things for no reason, I'll update the naming

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
nuno-vieira: It never scrolled to the end, I tried using DispatchQueue async, also using a fake view as an anchor, and it never scrolled accuratly to the end of the scroll view

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerFileAttachmentView.swift
martinmitrevski: 👍

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
martinmitrevski: good to extract a formatter already, and use it everywhere. If figma is not consistent, at least when it is, we know where to change it.

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
martinmitrevski: this one was about the localization, the other one is about the formatting

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
martinmitrevski: Welcome to SwiftUI :)

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
nuno-vieira: It can't be a computed property, because the files need to respect the order that they were added, independent of the type. If it is a computed property, we have no way to know which assets were added first

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
nuno-vieira: Numbers , times, etcc are not localisable. This is formatting, example from Apple: https://developer.apple.com/documentation/swift/duration/unitsformatstyle

martinmitrevski: not sure if you can do that with "8s" though. In any case, if you can, then sure, no need for localization.

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
martinmitrevski: hmm, but you're already computing the getter:
```
get: { [weak self] in
                guard let self else { return [] }
                return self.composerAssets.compactMap {
                    if case .addedFile(let url) = $0 { return url }
                    return nil
                }
            },
```

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
nuno-vieira: 8, s, m, h, sec, min etc usually are not localisable, but formattes can support localisation as well. Apple's time formatter supports localisation. In general, we should use Apple's timing formatters, but allow customers to override.

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
nuno-vieira: Fixed ✅

Sources/StreamChatSwiftUI/ChatComposer/Attachments/ComposerVideoAttachmentView.swift
martinmitrevski: yes, let's solve it with some standard formatter

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerView.swift
nuno-vieira: ✅

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
nuno-vieira: Yes but this is used for other things, the `composerAssets` is now the source of truth for the assets in the composer

nuno-vieira: I will check if we really need this, though. It feels like we don't really need this anymore 🤔

nuno-vieira: Done ✅

martinmitrevski (approved):
LGTM! We should move the images to the common UI module before merging though

PR #1201, opened 2026-02-13
V5: reactions overlay view refresh

laevandus:
### 🔗 Issue Links

Related: [IOS-1379](https://linear.app/stream/issue/IOS-1379)

### 🎯 Goal

Apply new design to the reactions overlay (timstamp, reaction picker) and rewrite the reactions overlay layout from absolute X/Y positioning to a VStack-based flow, making it easier to maintain and extend.

### 📝 Summary

- Rewrote `ReactionsOverlayView` to use a VStack layout with leading/trailing alignment instead of manual origin calculations
- Replaced pre-calculated sizes with measurement-based positioning via SwiftUI preference keys
- Moved user avatar inline into the message row HStack instead of absolute positioning
- Added support for bottom reactions, top reactions, and message date display within the overlay
- Replaced hardcoded spacing with design tokens
- Added optional `textColor` to `MessageDateView` for overlay styling
- Removed `ReactionsHStack`, `ChatMessage.reactionOffsetX`, and unused properties/helpers
- Simplified `ReactionsOverlayContainer` by removing wrapper and offset logic
- Updated snapshot tests to match the new layout

### 🛠 Implementation

The previous layout computed absolute X positions for every element using functions like `messageOriginX`, `messageActionsOriginX`, and `reactionOffsetX`, requiring tracking of screen width changes and iPad-specific diffs.

The new layout uses a single VStack with three sections: reactions picker, message + timestamp (with avatar), and message actions. Horizontal alignment is driven by `isRightAligned` with Spacers in each section's HStack. Vertical positioning uses `contentOffsetY` that animates from the message's original position to a computed fit-on-screen position, subtracting a measured `reactionsAreaOffset` so the message bubble (not the VStack top) aligns with its original position.

`ReactionsHStack` was removed since each section now manages its own alignment. `MessageDateView` gained an optional `textColor` parameter so the overlay can style the timestamp over the scrim background.

### 🎨 Showcase

Example

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

## Release Notes

* **New Features**
  * Message date text color can now be customized through factory options

* **Breaking Changes**
  * Removed `ReactionsHStack` public view component; custom implementations using this component will need to be updated


Sources/StreamChatSwiftUI/ChatMessageList/Reactions/ReactionsOverlayView.swift
```diff
@@ -240,6 +150,187 @@ public struct ReactionsOverlayView<Factory: ViewFactory>: View {
...
+                    )
+                }
+            )
+        }
+    }
+
+    // MARK: - Message + Timestamp (with avatar aligned to timestamp)
+
+    @ViewBuilder
+    private func messageAndTimestampView(reader: GeometryProxy) -> some View {
```
laevandus, line 210: In next PRs I should probably extract this piece and use it within MessageContainerView as well since the layout and logic should match. Currently message view spacings are a little bit off which needs fixing.

martinmitrevski (commented):
looks good codewise, but few edge cases we need to fix (shared screenshots)

Sources/StreamChatSwiftUI/ChatMessageList/Reactions/ReactionsOverlayView.swift
```diff
@@ -248,11 +338,11 @@ public struct ReactionsOverlayView<Factory: ViewFactory>: View {
             isFirst: messageDisplayInfo.isFirst,
             scrolledId: .constant(nil)
         )
-        // This is needed for the LinkDetectionTextView to work properly.
-        // TODO: This should be refactored on v5 so the TextView does not depend directly on the view model.
```
martinmitrevski, line 252: why do we delete this comment, it's still valid?

martinmitrevski (commented):
Fine with the code changes, but there are still few rough edges (shared on slack)

martinmitrevski (approved):

Sources/StreamChatSwiftUI/ChatMessageList/Reactions/ReactionsOverlayView.swift
laevandus: Added back, got lost when refactoring it

PR #1245, opened 2026-02-25
Typing indicator view redesign

laevandus:
### 🔗 Issue Links

Fixes: [IOS-1439](https://linear.app/stream/issue/IOS-1439/v5-typing-indicator-redesign)

  ### 🎯 Goal                                                                                                                                                                     
                             
  Redesign the typing indicator and move it from a bottom overlay into the message list as an inline element with user avatars.

  ### 📝 Summary

  - Redesigned `TypingIndicatorView` to show an avatar stack of typing users next to an incoming-style message bubble with animated dots
  - Moved the typing indicator from a bottom overlay into the message list (inline, flipped with the rest of the list)
  - Replaced `TypingIndicatorPlacement.bottomOverlay` with `.inline` and added `.automatic` (inline by default, falls back to navigation bar when scrolled up)
  - Changed the default placement to `.automatic`
  - Renamed `TypingIndicatorBottomView` / `makeTypingIndicatorBottomView` to `TypingIndicatorView` / `makeTypingIndicatorView`
  - Renamed the old `TypingIndicatorView` (animated dots) to `TypingIndicatorDotsView`

  ### 🛠 Implementation

  The typing indicator is now rendered as the first item inside the `LazyVStack` of the message list (flipped upside down like other messages). `TypingIndicatorView` composes an
  `AvatarStack` with a `TypingIndicatorDotsView` inside a `BubbleModifier`. The `TypingIndicatorPlacement` enum has three cases: `.navigationBar`, `.inline`, and `.automatic`. In
   `.automatic` mode the nav bar indicator only appears when `showScrollToLatestButton` is true (user scrolled away from the bottom).

### 🎨 Showcase

| Single user typing | Multiple users typing |
|---|---|
|  |  |

  ### 🧪 Manual Testing Notes

  - Verify typing indicator appears inline in the message list with user avatars
  - Test all three placement modes: `.navigationBar`, `.inline`, `.automatic`
  - For `.automatic`: verify the nav bar typing indicator appears only when scrolled up

### 🧪 Manual Testing Notes

  ### ☑️  Contributor Checklist


martinmitrevski (commented):
@laevandus shouldn't the message go up when the typing indicator is shown?

laevandus:
> @laevandus shouldn't the message go up when the typing indicator is shown? 

I am clarifying the layout logic

Sources/StreamChatSwiftUI/CommonViews/Avatars/AvatarStack.swift
```diff
@@ -42,12 +42,10 @@ public struct AvatarStack: View {
...
-        HStack(spacing: tokens.spacingSm - size) {
-            HStack(spacing: -tokens.spacingXs) {
-                ForEach(avatars.indices, id: \.self) { index in
-                    avatarView(url: avatars[index].url, initials: avatars[index].initials)
-                        .zIndex(Double(index))
-                }
+        HStack(spacing: -size / 2 + tokens.spacingXxs) {
+            ForEach(avatars.indices, id: \.self) { index in
+                avatarView(url: avatars[index].url, initials: avatars[index].initials)
+                    .zIndex(Double(index))
```
laevandus: Offset issue with the badge, it was too much to the left

martinmitrevski (approved):
LGTM! One note, good to check again the color of the typing indicator bubble's dots in dark mode, I think the visibility is not ideal (maybe also check with design)

laevandus:
> LGTM! One note, good to check again the color of the typing indicator bubble's dots in dark mode, I think the visibility is not ideal (maybe also check with design)

Animation used 0.1 alpha while the new design is using 0.25. I'll just in case clarify with design as well. Fixed e2e test as well.

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelView.swift
```diff
@@ -95,9 +95,9 @@ public struct ChatChannelView<Factory: ViewFactory>: View, KeyboardReadable {
                                 .dismissKeyboardOnTap(enabled: keyboardShown) {
                                     hideComposerCommandsAndAttachmentsPicker()
                                 }
-                            if viewModel.shouldShowTypingIndicator {
-                                factory.makeTypingIndicatorBottomView(
-                                    options: TypingIndicatorBottomViewOptions(
+                            if viewModel.shouldShowInlineTypingIndicator {
+                                factory.makeTypingIndicatorView(
+                                    options: TypingIndicatorViewOptions(
```
nuno-vieira, line 100: The factory method is used here, but not in the Header 🤔 Shouldn't we call it on the header view as well?

Sources/StreamChatSwiftUI/ChatChannel/ChannelHeader/ChatChannelHeaderViewModifier.swift
```diff
@@ -154,7 +152,7 @@ public struct ChannelTitleView: View {
 
             if shouldShowTypingIndicator {
                 HStack {
-                    TypingIndicatorView()
+                    TypingIndicatorDotsView()
```
nuno-vieira: Shouldn't this come from the Factory as well?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
```diff
@@ -126,7 +126,7 @@ public struct ChatChannelListItem<Factory: ViewFactory>: View {
                     .foregroundColor(Color(colors.accentError))
                     .lineLimit(1)
             } else if channel.shouldShowTypingIndicator {
-                TypingIndicatorView()
+                TypingIndicatorDotsView()
```
nuno-vieira: Shouldn't this also come from the factory?

nuno-vieira (approved):
LGTM! Just add some comments about not having all of the views in the factory. Since most likely customers would like to replace the view in one place, and reflect it everywhere. But we can add the factory methods later.

Sources/StreamChatSwiftUI/ChatChannel/ChatChannelView.swift
laevandus: I will add factory for this one. It used be in a way where there was additional wrapper view TypingIndicatorBottomView which wrapped TypingIndicatorView. Only the bottom view used factory. But let me change this around and add a factory for the header side as well.

laevandus: Now we have
```
 public func makeInlineTypingIndicatorView(
        options: TypingIndicatorViewOptions
    ) -> some View {
        let users = Array(options.channel.currentlyTypingUsersFiltered(currentUserId: options.currentUserId))
        let typingText = options.channel.typingIndicatorString(currentUserId: options.currentUserId)
        return TypingIndicatorView(users: users, typingText: typingText)
    }
    
    public func makeSubtitleTypingIndicatorView(
        options: SubtitleTypingIndicatorViewOptions
    ) -> some View {
        SubtitleTypingIndicatorView(channel: options.channel)
    }
```

PR #1273, opened 2026-03-09
Merge develop to v5

testableapple:


StreamChatSwiftUI.xcodeproj/project.pbxproj
```diff
@@ -1544,8 +1544,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/GetStream/stream-chat-swift.git";
 			requirement = {
-				kind = revision;
-				revision = 5fa05f4151d4a0a680d0f86aef1d7a393352f983;
+				branch = v5;
+				kind = branch;
```
martinmitrevski: shouldn't we have a commit here?

martinmitrevski (commented):
The snapshots are with the old designs (maybe they are not used/deleted in v5)?

StreamChatSwiftUI.xcodeproj/project.pbxproj
testableapple: `develop` is merged to `v5` on LLC, so i don't think we have to use commit

martinmitrevski: I think we agreed to use a commit, so we don't break the build when things are changed on the LLC

martinmitrevski (approved):

StreamChatSwiftUI.xcodeproj/project.pbxproj
testableapple: reverted 👍

PR #1303, opened 2026-03-17
v5: Message List Polish

nuno-vieira:
### 🔗 Issue Links

[IOS-1492](https://linear.app/stream/issue/IOS-1492/v5-message-list-polish)

### 🎯 Goal

Polish the message list, channel list, and related views to match the v5 Figma design system specs, fix interaction issues, and refactor search into a reusable modifier.

### 📝 Summary

- Fix background colors across multiple views (message list, channel list, search results, polls, reactions, info screens) to use the correct design tokens (`backgroundElevationElevation0`, `backgroundCoreApp`)
- Fix Giphy badge sizing/styling and Giphy preview height with proper aspect ratio (4/3) and minimum height
- Fix date indicator pill spacing and padding to match Figma specs
- Fix long press gesture to work on the full message cell (not just the bubble) and reduce activation time using `simultaneousGesture`
- Fix search bar styling (colors, icons, spacing) to match Figma specs
- Add bottom spacing after pinned message background to separate it from the next message
- Introduce `Styles.makeSearchableModifier()` as a reusable search modifier, removing `makeChannelListTopView` from `ViewFactory`

### 🛠 Implementation

**Background colors**: Updated `DefaultViewFactory.makeMessageListBackground` and multiple views (`SearchResultsView`, `PollAllOptionsView`, `PollCommentsView`, `PollResultsView`, `ReactionsDetailView`, `MoreReactionsView`, `ReactionsOverlayView`, etc.) to use the correct elevation tokens from the design system.

**Giphy attachments**: Reworked `GiphyBadgeView` to use the correct sizing tokens (`iconSizeXs`, `spacingXs`, `radiusLg`). Reworked `LazyGiphyView` to use a `Color.clear` container with `aspectRatio(4/3)` and `minHeight(192)` as the layout driver, with the image overlaid and clipped.

**Message gestures**: Added `.contentShape(Rectangle())` to `MessageContainerView` in `MessageItemView` so the long press hit area covers the full cell. Replaced `.onLongPressGesture` with `.simultaneousGesture(LongPressGesture(minimumDuration: 0.35))` to resolve gesture disambiguation with double-tap and swipe gestures.

**Search bar**: Updated `SearchBar` to match the Figma spec (border instead of filled background, correct icon colors, font, and spacing).

**Pinned messages**: Added conditional bottom padding (`tokens.spacingXs`) to `MessageItemView` when a message is pinned.

**Search refactoring**: Created `SearchableModifierOptions` and `DefaultSearchableModifier` in `Styles.swift`. The default uses the SDK's styled `SearchBar` for full appearance control and snapshot test compatibility. Removed `makeChannelListTopView` / `ChannelListTopViewOptions` / `ChannelListTopViewType` from the `ViewFactory` protocol. Updated `ChatChannelListContentView` and `AddUsersView` to use the new modifier. Updated `RedactedLoadingView` to remove the now-deleted factory method.

### 🧪 Manual Testing Notes

- Open the channel list and verify the search bar appears styled correctly and scrolls with content
- Search for channels and verify results render with the correct background
- Open a channel and verify message list background, date dividers, and pinned message spacing
- Long press on a message (outside the bubble area) and verify the actions overlay opens quickly (~0.3s)
- Send a Giphy and verify the badge and preview sizing
- Open the Add Users sheet and verify the search bar works
- Verify poll views, reactions detail, and other info screens have correct backgrounds

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
```diff
@@ -232,29 +230,44 @@ public struct ChatChannelListContentView<Factory: ViewFactory>: View {
...
-                    onItemAppear: { index in
-                        viewModel.checkTabBarAppearance()
-                        viewModel.checkForChannels(index: index)
-                    },
-                    channelDestination: viewFactory.makeChannelDestination(options: ChannelDestinationOptions()),
-                    trailingSwipeRightButtonTapped: viewModel.onDeleteTapped(channel:),
-                    trailingSwipeLeftButtonTapped: viewModel.onMoreTapped(channel:),
-                    leadingSwipeButtonTapped: { _ in /* No leading button by default. */ }
-                )
+                ScrollViewReader { scrollView in
```
martinmitrevski: this doesn't make sense - the `ChannelList` already has a scroll view reader and scroll view inside, and we're passing `scrollable` false to disable that. Why have it twice, it should be either here or there. The channel list should be scrollable as a component imo.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
```diff
@@ -232,29 +230,44 @@ public struct ChatChannelListContentView<Factory: ViewFactory>: View {
...
+                            channelDestination: viewFactory.makeChannelDestination(options: ChannelDestinationOptions()),
+                            trailingSwipeRightButtonTapped: viewModel.onDeleteTapped(channel:),
+                            trailingSwipeLeftButtonTapped: viewModel.onMoreTapped(channel:),
+                            leadingSwipeButtonTapped: { _ in }
+                        )
+                    }
+                    .onChange(of: viewModel.scrolledChannelId) { newValue in
+                        if let newValue {
+                            withAnimation {
+                                scrollView.scrollTo(newValue, anchor: .bottom)
```
martinmitrevski: or we can handle this inside the channel list

Sources/StreamChatSwiftUI/ChatMessageList/GiphyAttachmentView.swift
```diff
@@ -124,27 +124,34 @@ struct LazyGiphyView: View {
...
-                } else {
-                    state.image
-                }
-            } else if state.error != nil {
-                Color(.secondarySystemBackground)
-            } else {
-                ZStack {
-                    Color(.secondarySystemBackground)
-                    ProgressView()
+        Color.clear
```
martinmitrevski: Feels like a hack

Sources/StreamChatSwiftUI/ChatMessageList/GiphyAttachmentView.swift
```diff
@@ -124,27 +124,34 @@ struct LazyGiphyView: View {
...
+            .aspectRatio(4 / 3, contentMode: .fit)
```
martinmitrevski: why do we hardcode the ratio? Not all gifs have the same ratio

Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
```diff
@@ -106,14 +106,18 @@ public struct MessageItemView<Factory: ViewFactory>: View {
...
+                .contentShape(Rectangle())
                 .onTapGesture(count: 2) {
                     if messageViewModel.isDoubleTapOverlayEnabled {
                         handleGestureForMessage(showsMessageActions: true)
                     }
                 }
-                .onLongPressGesture(perform: {
-                    handleGestureForMessage(showsMessageActions: true)
-                })
+                .simultaneousGesture(
```
martinmitrevski, line 116: some of this stuff breaks the tap to dismiss gesture on the message item

Sources/StreamChatSwiftUI/CommonViews/SearchBar.swift
```diff
@@ -48,15 +48,14 @@ struct SearchBar: View, KeyboardReadable {
                     }
                 }
             }
-            .padding(.horizontal, tokens.spacingSm)
+            .padding(.horizontal, tokens.spacingMd)
             .padding(.vertical, tokens.spacingSm)
-            .background(Color(colors.backgroundElevationElevation1))
+            .frame(minHeight: 48)
```
martinmitrevski: why minHeight?

Sources/StreamChatSwiftUI/ViewFactory/ViewFactory.swift
```diff
@@ -73,12 +73,6 @@ import SwiftUI
     /// - Returns: View displayed in the leading swipe area of a channel item.
     func makeLeadingSwipeActionsView(options: LeadingSwipeActionsViewOptions) -> LeadingSwipeActionsViewType
 
-    associatedtype ChannelListTopViewType: View
-    /// Creates the view shown at the top of the channel list. Suitable for search bar.
-    /// - Parameter options: the options for creating the channel list top view.
-    /// - Returns: view shown above the channel list.
-    func makeChannelListTopView(options: ChannelListTopViewOptions) -> ChannelListTopViewType
```
martinmitrevski, line 80: why do we remove this? How are we going to add a custom view at the top of the list?

Sources/StreamChatSwiftUI/Styles.swift
```diff
@@ -312,6 +323,50 @@ public struct LiquidGlassScrollToBottomButtonModifier: ViewModifier {
...
+
+    public init(searchText: Binding<String>, prompt: String) {
+        _searchText = searchText
+        self.prompt = prompt
+    }
+
+    public func body(content: Content) -> some View {
+        if #available(iOS 17.0, *) {
+            content
+                .searchable(
```
martinmitrevski, line 356: the problem with this is that it doesn't look like the designs now, is there some way to customize it?

martinmitrevski (changes requested):
Several issues with this PR

Sources/StreamChatSwiftUI/ChatMessageList/GiphyAttachmentView.swift
nuno-vieira: It is following the figma specs, giphies alwyas have this ratio and a minHeight. The contentMode will crop the image, so even if the original giphy has a different aspect ration it should be okay

Sources/StreamChatSwiftUI/ChatMessageList/GiphyAttachmentView.swift
nuno-vieira: Since we are using a LazyImage, the aspect ratio was not working, so a container is used instead. This was actualy the origin of the bug

Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
nuno-vieira: In what scenario?

Sources/StreamChatSwiftUI/CommonViews/SearchBar.swift
nuno-vieira: I would assume it was taken from Figma, at least visually it looked similar to Figma design

Sources/StreamChatSwiftUI/ViewFactory/ViewFactory.swift
nuno-vieira: At the moment, this would be empty. But a modifier is much more flexible for these kinds of things than putting an explicit view at the top. By using the channel list modifier, you can easily insert a view as the top

nuno-vieira: But I can add this one back since we also have the footer, so it does not make sense to remove it

Sources/StreamChatSwiftUI/Styles.swift
nuno-vieira: I tried to customise it, but there is no way to do it at the moment. Either way, most customers will just want the default search UI, and not a custom one. Besides that, this one adapts to Liquid glass:

Image

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
nuno-vieira: No idea we had a "scrollable" property. I think maybe this was added previously when AI refactored the old SearchBar to be scrollable. So instead of moving the SearchBar Inside ChannelList, it did this. Will remove this then

nuno-vieira: Fixed

Sources/StreamChatSwiftUI/ChatMessageList/GiphyAttachmentView.swift
nuno-vieira: Added a maxHeight instead, to avoid this hack and avoid the aspect ratio

Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
nuno-vieira: Fixed it

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
martinmitrevski: doesn't seem removed, did you push?

Sources/StreamChatSwiftUI/CommonViews/SearchBar.swift
martinmitrevski: not saying about the number, but why is it `minHeight`? It should just be `height` imo.

Sources/StreamChatSwiftUI/ViewFactory/ViewFactory.swift
martinmitrevski: I think we should bring it back and return empty view by default. Many customers were using this, for example for pinned channels.

Sources/StreamChatSwiftUI/Styles.swift
martinmitrevski: can you check with Jurgen if that's ok? btw, this is the uikit app

Sources/StreamChatSwiftUI/ViewFactory/ViewFactory.swift
nuno-vieira: Fixed

Sources/StreamChatSwiftUI/CommonViews/SearchBar.swift
nuno-vieira: Fixed

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
nuno-vieira: Fixed

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
```diff
@@ -214,11 +216,11 @@ public struct ChatChannelListContentView<Factory: ViewFactory>: View {
 
     public var body: some View {
         VStack(spacing: 0) {
-            viewFactory.makeChannelListTopView(
-                options: ChannelListTopViewOptions(searchText: $viewModel.searchText)
-            )
-
             if viewModel.isSearching {
+                viewFactory.makeChannelListTopView(
```
martinmitrevski: why have this twice, we can move it up for both cases?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
```diff
@@ -232,12 +234,16 @@ public struct ChatChannelListContentView<Factory: ViewFactory>: View {
...
+                    options: ChannelListTopViewOptions()
+                )
+
                 ChannelList(
                     factory: viewFactory,
                     channels: viewModel.channels,
                     selectedChannel: $viewModel.selectedChannel,
                     swipedChannelId: $viewModel.swipedChannelId,
-                    scrolledChannelId: $viewModel.scrolledChannelId,
```
martinmitrevski, line 240: why aren't we passing the `viewModel.scrolledChannelId` anymore?

Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
martinmitrevski: works well now

Sources/StreamChatSwiftUI/CommonViews/LoadingView.swift
```diff
@@ -22,19 +22,11 @@ public struct RedactedLoadingView<Factory: ViewFactory>: View {
 
     public var body: some View {
         ScrollView {
-            VStack(spacing: 0) {
-                factory.makeChannelListTopView(
```
martinmitrevski, line 26: probably we need to bring this back?

martinmitrevski (commented):
few more things to check before merging, also there seems to be some conflicts with v5

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListView.swift
nuno-vieira: Fixed

Sources/StreamChatSwiftUI/CommonViews/LoadingView.swift
nuno-vieira: Fixed

martinmitrevski (approved):

PR #1330, opened 2026-03-24
Fix: Single image/video attachment renders sharp tail corner when no caption

nuno-vieira:
### 🔗 Issue Links

[IOS-1518](https://linear.app/stream/issue/IOS-1518)

### 🎯 Goal

Single image/video attachments without a caption should render a sharp corner on the tail side (bottom-left for incoming, bottom-right for outgoing) to visually indicate the absence of a caption, matching the Figma design spec.

### 📝 Summary

- Add per-corner clipping support to `MessageMediaAttachmentContentView` via a new `corners: UIRectCorner?` parameter
- Detect single-media-without-caption in `MessageMediaAttachmentsContainerView` and apply bubble-matching corners with a border overlay
- Pass `isFirst` from the factory to the container so the sharp corner only applies to first-in-group messages
- Add 6 new snapshot tests covering outgoing/incoming, first/non-first in group, portrait orientation, video, and captioned variants
- Re-record 5 existing snapshot references affected by the corner change
- Update CHANGELOG

### 🧪 Manual Testing Notes

1. Send a single image without any text → the tail corner should be sharp
2. Send a single video without any text → same sharp tail corner
3. Send a single image with a caption → all corners should be uniformly rounded inside the bubble
4. Send multiple images → gallery cells should have uniform rounded corners
5. Verify both incoming and outgoing messages display the correct sharp corner side

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Single image and video attachments without captions now render with the correct sharp tail corners and matching border/stroke for a polished bubble appearance.

* **Tests**
  * Added snapshot tests covering single-media-without-caption scenarios (incoming/outgoing, first/non-first in group, multiple aspect ratios, and video).

* **Documentation**
  * Changelog updated to reflect the fix.


Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
```diff
@@ -114,12 +119,54 @@ public struct MessageMediaAttachmentsContainerView<Factory: ViewFactory>: View {
...
     }
 
+    @ViewBuilder
     private func singleItemLayout(
         _ item: MediaAttachment,
         width: CGFloat,
         height: CGFloat
     ) -> some View {
-        mediaCell(item, width: width, height: height, index: 0)
+        if isSingleMediaWithoutCaption {
```
martinmitrevski: we shouldn't use if-else here, it's the same code. We should just pass the corners based on this bool.

nuno-vieira: Done

martinmitrevski (approved):

PR #1332, opened 2026-03-25
[v5] Fix direct message channel avatars and avatar factory methods

laevandus:
### 🔗 Issue Links

Fixes [IOS-1551](https://linear.app/stream/issue/IOS-1551/v5-channelavatar-direct-message-presentation)

### 🎯 Goal

- Show user avatar in 1:1 DMs with presence

### 📝 Summary

- Fix 1:1 DM avatars and show presence if the other party is online
- Use factory methods for creating avatars everywhere
- Make it easier to reconfigure presence indicators in view factories by exposing `indicator` argument
- Refactored tests for channel avatars

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  Before   |  After  |

### 🧪 Manual Testing Notes

1. 1:1 DMs show other participant's avatar and presence if online
2. 3 and more participant DMs show stacked avatars
3. Groups show always stacked avatars (if the channel does not have custom avatar set)

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed rendering of avatars and presence indicators in 1:1 direct messages.

* **Refactor**
  * Standardized avatar rendering across the app to improve consistency and enable configurable avatar indicators and borders.


Sources/StreamChatSwiftUI/CommonViews/Avatars/ChannelAvatar.swift
```diff
@@ -26,22 +27,22 @@ public struct ChannelAvatar: View {
...
-    ///   - showsIndicator: A Boolean value that indicates whether to show the
-    ///     online status for direct message channels. Defaults to `false`.
+    ///   - indicator: The presence indicator to display (e.g. `.online`, `.none`).
     ///   - showsBorder: A Boolean value that indicates whether to show a circular
     ///     border around the avatar. Defaults to `true`.
     public init(
         channel: ChatChannel,
         size: CGFloat,
-        showsIndicator: Bool = false,
+        indicator: AvatarIndicator = .none,
```
laevandus, line 36: I changed it around because then it is more flexible to reconfigure presence logic when overriding avatar factory methods.

Sources/StreamChatSwiftUI/CommonViews/Avatars/ChannelAvatar.swift
```diff
@@ -60,6 +61,8 @@ public struct ChannelAvatar: View {
     ///     used for the stacked layout when `url` is `nil`.
     ///   - memberCount: The total number of members in the channel, used to
     ///     compute the overflow badge count.
+    ///   - directMessageChannel: A Boolean value that indicates whether the
```
nuno-vieira, line 64: Can't this be calculated from the `ChatChannel` that is already in this view? 🤔

StreamChatSwiftUITests/Tests/CommonViews/Avatars/ChannelAvatar_Tests.swift
```diff
@@ -12,179 +12,206 @@ import XCTest
...
+
         // Then
         AssertSnapshot(view, size: size)
     }
-    
+
     func test_channelAvatar_stackedPlaceholders_withOnlineIndicator() {
-        // Given — stacked layout with presence indicator
+        // Given — DM channel with online member shows indicator
+        let channel = mockGroupChannel(memberCount: 2, activeMembers: 2)
```
nuno-vieira, line 120: I think that  Group channel with 2 members, is different than a true DM channel with 2 members 🤔 Is the avatar supposed to look like the same in both scenarios? Or only in a true DM Channel?

Sources/StreamChatSwiftUI/CommonViews/Avatars/ChannelAvatar.swift
martinmitrevski: we don't keep the `ChatChannel` in the view (and we shouldn't). When the channel param is provided in the init, that's already like that in the other init above.

martinmitrevski (approved):

StreamChatSwiftUITests/Tests/CommonViews/Avatars/ChannelAvatar_Tests.swift
laevandus: Only true 1:1 DM channels get the large participant image, groups always use stacked ones.

PR #1349, opened 2026-03-31
Update deleted message design

nuno-vieira:
### 🔗 Issue Links

Fixes [IOS-1505](https://linear.app/stream/issue/IOS-1505/v5-qa-deleted-message-does-not-match-figma)

### 🎯 Goal

The deleted message UI did not match the Figma design. This PR updates the `DeletedMessageView` to match the design spec, including adding the nosign icon, using the correct text color, and showing the delivery status below deleted messages.

### 📝 Summary

- Add a "nosign" icon (16×16) before the "Message deleted" text in `DeletedMessageView`
- Use the standard message text color (`chatTextOutgoing`/`chatTextIncoming`) instead of `textLowEmphasis`
- Remove the "Only visible to you" footer from `DeletedMessageView` (delivery status is now handled by the parent container)
- Show delivery status (read indicator + timestamp) for deleted messages in `MessageContainerView`
- Update `textColor(for:)` helper to no longer special-case deleted messages

### 🛠 Implementation

- **`DeletedMessageView`**: Simplified the view to an `HStack` with the nosign SF Symbol and the placeholder text. Removed the `chatClient`, `images`, and `utils` dependencies since they are no longer needed. Uses design tokens (`tokens.spacingXxs`, `tokens.iconSizeSm`) for spacing and icon sizing.
- **`MessageContainerView`**: Removed the `!message.isDeleted` condition gating the delivery status view, so deleted messages now show read indicators and timestamps like regular messages.
- **`MessageListHelperViews`**: Removed the `isDeleted` branch in `textColor(for:)` that returned `textLowEmphasis`, so deleted messages render with the same text color as normal messages.

### 🧪 Manual Testing Notes

- Send a message, then delete it. The deleted message should display the nosign icon + "Message deleted" text in the standard text color, with the delivery status (checkmark + timestamp) visible below the bubble.
- Verify both outgoing and incoming deleted messages render correctly.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified deleted message appearance and consolidated styling.
  * Deleted messages now show delivery/read status when "show all info" is enabled.
  * Standardized text color handling for deleted messages.

* **Tests**
  * Added snapshot and updated delivery-status tests covering deleted-message behavior.

* **Changelog**
  * Added changelog entry summarizing the deleted-message design update.


Sources/StreamChatSwiftUI/ChatMessageList/DeletedMessageView.swift
```diff
@@ -7,59 +7,33 @@ import SwiftUI
...
     @Injected(\.colors) private var colors
-    @Injected(\.utils) private var utils
-
-    private var dateFormatter: DateFormatter {
-        utils.dateFormatter
-    }
-    
-    private var deletedMessageVisibility: ChatClientConfig.DeletedMessageVisibility {
-        chatClient.config.deletedMessagesVisibility
-    }
```
nuno-vieira, line 22: @martinmitrevski Do we want to keep supporting this? These have caused a lot of trouble in the past. Having deleted message logic client side only is very problematic since it can break pagination, etc... Soft deleting, Hard deleting, or deleting only for your self should be a backend feature

martinmitrevski (approved):

Sources/StreamChatSwiftUI/ChatMessageList/DeletedMessageView.swift
martinmitrevski: nope, we don't need this anymore

laevandus (approved):

PR #1359, opened 2026-04-02
Fix media gallery attachment not opening the correct item on first tap

nuno-vieira:
### 🔗 Issue Links

[IOS-1503](https://linear.app/stream/issue/IOS-1503/v5-qa-tapping-non-first-attachment-opens-the-first-attachment-instead)

### 🎯 Goal

Fix the bug where tapping on a non-first media attachment in a message always opens the first attachment on the initial tap.

### 📝 Summary

- Use `onChange(of: selectedIndex)` to trigger gallery presentation, ensuring the selected index is committed before `fullScreenCover` evaluates its content closure.
- Handle the edge case where the same index is tapped by directly toggling `galleryShown`.

### 🛠 Implementation

The root cause was a SwiftUI timing issue: setting `selectedIndex` and `galleryShown` in the same tap handler meant `fullScreenCover`'s content closure could evaluate before SwiftUI committed the new `selectedIndex`, resulting in the default value of `0` being used on the first tap.

The fix separates the state update from the presentation trigger by using `onChange(of: selectedIndex)` to set `galleryShown = true` only after SwiftUI has committed the new index value. For the edge case where the user taps the already-selected index, `galleryShown` is set directly since no index change is needed.

### 🧪 Manual Testing Notes

1. Open a channel with messages containing multiple media attachments.
2. Tap on the second, third, or any non-first attachment.
3. Verify the media viewer opens on the correct attachment on the first tap.
4. Dismiss and tap another attachment to confirm consistent behavior.

### ☑️ Contributor Checklist

## Release Notes

* **Bug Fixes**
  * Fixed an issue where tapping a non-first media attachment would incorrectly open the first item instead of the selected attachment.


Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
```diff
@@ -80,9 +80,10 @@ public struct MessageMediaAttachmentsContainerView<Factory: ViewFactory>: View {
 
     public var body: some View {
         galleryGrid
-            .fullScreenCover(isPresented: $galleryShown, onDismiss: {
-                selectedIndex = 0
-            }) {
+            .onChange(of: selectedIndex) { _ in
+                galleryShown = true
```
martinmitrevski, line 84: hmm, is this safe? If it changes from some value to `nil`, this would be triggered again. We should do a nil check comparison.

nuno-vieira: What do you mean? SelectedIndex can't be nil

martinmitrevski: ah ok, nvm, I thought it's another view

martinmitrevski (approved):

PR #1374, opened 2026-04-06
Fix scrolling long messages in reactions overlay

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1583/v5-fix-scrolling-long-messages-in-reactions-overlay.

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interaction handling for message previews in scroll-view-based layouts, allowing users to tap and interact with messages that were previously unresponsive.


Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
```diff
@@ -107,7 +107,7 @@ public struct MessageItemView<Factory: ViewFactory>: View {
                     }
                 )
                 .contentShape(Rectangle())
-                .allowsHitTesting(!shownAsPreview)
+                .allowsHitTesting(!shownAsPreview || (messageViewModel.usesScrollView))
```
nuno-vieira, line 110: The problem here is that the space between the reactions and messages will still not close the reactions. Are we fine with this?

martinmitrevski: yes, it's fine for this case

laevandus: There is another side-effect which is download buttons, reactions etc work when message is long, but does not for messages not using scroll view. At some point we should implement the latest design where long messages go underneath message actions and are clipped if very long (that was added after I touched this view so it has the old scroll view behaviour).

But yeah, I guess for non beta release this fix is OK and later we should apply the very latest design update.

nuno-vieira (approved):

PR #1376, opened 2026-04-06
Fix view button in a thread not working the first time

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1585/v5-fix-view-button-in-a-thread-not-working-the-first-time.

### 🎯 Goal

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed navigation behavior where jumping to a message in a channel would unexpectedly open associated threads. Messages now display correctly without triggering automatic thread navigation.


Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -265,8 +265,15 @@ import SwiftUI
...
+        let wasInThread = threadMessageShown
         threadMessageShown = false
-        _ = jumpToMessage(messageId: messageId)
+        if wasInThread {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) { [weak self] in
+                _ = self?.jumpToMessage(messageId: messageId)
+            }
+        } else {
+            _ = jumpToMessage(messageId: messageId)
+        }
```
laevandus: I briefly looked into this as well and seems like the alternative is to change `jumpToMessage`:
```
let message = channelController.dataStore.message(id: baseId)
                if let parentMessageId = message?.parentMessageId, !isMessageThread {
                    let parentMessage = channelController.dataStore.message(id: parentMessageId)
                    threadMessage = parentMessage
                    threadMessageShown = true
                    messageCachingUtils.jumpToReplyId = message?.messageId
                    return false
                }
```
Here the `baseId` is the message sent to the channel. Issue is that here jumping destination is changed to its parent message although we really want to jump to the id specified by the `baseId`.
And this code path is getting hit only because `self.messages` are not including the message which was just posted to the channel. Ideally we should hit the `loadPageAroundMessageId` a little bit later in that function.

I am not sure when this flow stopped working because when I added it, it did not have this problem. I guess we changed some configs, not sure.

I wonder if this fix works if I jump to a really old message first.

martinmitrevski: yes, this solution is better, updated it

nuno-vieira (approved):

PR #1369, opened 2026-04-06
Fix file attachments border

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1580/v5-fix-file-attachments-border.

### 🎯 Goal

_Describe why we are making this change._

### 📝 Summary

_Provide bullet points with the most important changes in the codebase._

### 🛠 Implementation

_Provide a detailed description of the implementation and explain your decisions if you find them relevant._

### 🎨 Showcase

_Add relevant screenshots and/or videos/gifs to easily see what this PR changes, if applicable._

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

_Explain how this change can be tested manually, if applicable._

### ☑️ Contributor Checklist

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] This change should be manually QAed
- [ ] Changelog is updated with client-facing changes
- [ ] Changelog is updated with new localization keys
- [ ] New code is covered by unit tests
- [ ] Documentation has been updated in the `docs-content` repo

## Summary by CodeRabbit

* **Style**
  * Improved file attachment visuals: attachments render with consistent rounded corners that respect message alignment and layout direction.
  * Updated attachment borders and clipping for a cleaner, more consistent appearance across message lists.
  * Enhanced behavior for first attachments in messages to ensure correct border and corner rendering when messages contain a single attachment or no text.

PR #1388, opened 2026-04-07
Fix voice recording attachment container when quoting a message

nuno-vieira:
### 🔗 Issue Links

[IOS-1590](https://linear.app/stream/issue/IOS-1590/v5-voice-message-quoted-reply-wrong-design)

### 🎯 Goal

Fix the voice recording attachment layout when the message is a quoted reply without a text caption. Per the design spec, the voice player should render flat inside the message bubble without its own container (background + border).

### 📝 Summary

- Add a `VoiceRecordingContainerModifier` to `MessageAttachmentsBubbleConfiguration` that conditionally applies the attachment container styling (padding, background, and rounded border).
- When the voice recording is a quoted reply without a text caption, the container is omitted so the player sits flat inside the message bubble.
- Add snapshot tests covering voice recording as a quoted reply (with and without text caption), for both incoming and outgoing messages.

### 🛠 Implementation

A new `VoiceRecordingContainerModifier` `ViewModifier` is nested inside `MessageAttachmentsBubbleConfiguration`. It checks whether the message has a `quotedMessage` and an empty `text` — if both are true, it skips the background and rounded border, letting the player blend into the parent bubble. Otherwise it applies the existing container styling (padding + background color + rounded border). `VoiceRecordingContainerView` now uses this modifier instead of applying the styling directly.

### 🧪 Manual Testing Notes

1. Send a voice recording as a reply to another message (no text caption) — the voice player should appear without a separate container.
2. Send a voice recording as a reply with a text caption — the voice player should still show its container.
3. Send a standalone voice recording (no reply, no text) — should render as before.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed rendering/styling of voice recording attachment containers when quoting a message so they display consistently (padding, background, and rounded border applied only when appropriate).

* **Tests**
  * Added snapshot tests covering voice recording messages with quoted replies for both sender directions and with/without quoted text.


StreamChatSwiftUITests/Tests/ChatChannel/__Snapshots__/MessageView_Tests/test_messageViewVoiceRecordingFromMe_snapshot.default-light.png
martinmitrevski, line 1: hmm, the border here is not right (see the bottom right corner). We had a similar issue with the file attachments here: https://github.com/GetStream/stream-chat-swiftui/pull/1369. Maybe good to apply the same fix.

martinmitrevski (approved):
LGTM for the container, just check the border

StreamChatSwiftUITests/Tests/ChatChannel/__Snapshots__/MessageView_Tests/test_messageViewVoiceRecordingFromMe_snapshot.default-light.png
nuno-vieira: Done ✅ I've also extracted a modifer to share the logic

PR #1400, opened 2026-04-09
Remove MessageViewModel environment workaround for translated text

laevandus:
### 🔗 Issue Links

Fixes [IOS-1022](https://linear.app/stream/issue/IOS-1022/v5-swiftui-swiftui-get-rid-of-the-environment-workaround)

### 🎯 Goal

Fixes the environment workaround for making translated text working

### 📝 Summary

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

Test various flows with translated text. Editing, toggling show/hide original text.

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/ChatMessageList/MessageView.swift
```diff
@@ -16,17 +16,20 @@ public struct MessageView<Factory: ViewFactory>: View {
     public var message: ChatMessage
     public var contentWidth: CGFloat
     public var isFirst: Bool
+    public let text: String
```
nuno-vieira: Maybe we can rename to messageText? Or messageTextContent?

martinmitrevski (commented):
I find it a bit confusing API wise to pass both message (which has message.text) and a text property. Maybe we can explore other options?

Sources/StreamChatSwiftUI/ChatMessageList/MessageView.swift
martinmitrevski: I think having this extra property is confusing, we should explore other options

nuno-vieira: The only possible option is to pass the view model in the constructor, there aren't many options here 🤔

laevandus: I gonna try to explore more ideas.
Other best idea I had was to go full attributed string here so view model creates it, we pass it down to other views. API wise better since it gives better understanding why we have this special argument for text when we also pass down message. Also we could get rid of the LinkDetectionTextView all together and make the text rendering more dumber. Only issue is that AttributedString is iOS 15 so a tiny wrapper must be used for avoiding converting AttributedString to NSAttributedString and then back (Text wants AttributedString). I can think about this approach instead.

martinmitrevski: Link detection text view had some logic around opening links from markdown, that some customers used. I wouldn't delete that tbh. 
Anyway, both options work for me.

PR #1408, opened 2026-04-10
Redesign attachment uploading and error states

nuno-vieira:
### 🔗 Issue Links

[IOS-1610](https://linear.app/stream/issue/IOS-1610/v5-qa-attachments-missing-new-error-and-loading-states)

### 🎯 Goal

Redesign the uploading progress and error state indicators for media and file attachments to match the updated Figma design system specs.

### 📝 Summary

- Replace the old percentage progress overlay with a new `LoadingSpinnerView` that supports both indeterminate (loading) and determinate (upload progress) modes
- Add a `RetryBadgeView` for upload and thumbnail load failures on media attachments, with tap-to-retry support
- Show inline upload progress (spinner + byte count) and failure state (warning icon + retry link) for file attachments inside `FileAttachmentDisplayView`
- Remove `PercentageProgressView` and `AttachmentDownloadingStateView` (no longer used)
- Stop shimmer animation when a media thumbnail fails to load

### 🛠 Implementation

**LoadingSpinnerView** now accepts an optional `progress: Double?`. When `nil`, it renders the existing indeterminate spinning animation. When set, it draws a determinate arc proportional to the value (0–1). Predefined sizes include a new `.medium` (24pt) tier.

**RetryBadgeView** is a new 32×32 red circular badge with a white border and a retry icon, shown when an upload or thumbnail load fails. Tapping it on media attachments retries the thumbnail load; tapping it on file attachments triggers `resendMessage()`.

**MessageMediaAttachmentContentView** was refactored to manage overlay states internally — the `withUploadingStateIndicator` modifier is no longer applied at the container level for media. The video play icon is now hidden during uploads and errors. A loading-error overlay stops the shimmer and shows the retry badge.

**FileAttachmentDisplayView** was extended with optional `uploadingState` and `onRetry` parameters so the same view handles the normal, uploading, and failed layouts without duplicating the file row structure. `FileAttachmentView` delegates to it, passing upload state and a retry closure.

### 🧪 Manual Testing Notes

- Send a message with image/video/file attachments on a slow or failing network
- Verify the spinner shows determinate progress during upload
- Simulate upload failure and confirm the retry badge appears; tap to retry
- Verify file attachments show inline progress text and "Retry upload" link on failure

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added inline upload progress and retry functionality for file attachments
  * Introduced retry badge for failed uploads

* **Improvements**
  * Redesigned attachment uploading progress and error state indicators

* **Localization**
  * Added new strings for upload status messages


Sources/StreamChatSwiftUI/ChatMessageList/FileAttachmentView.swift
```diff
@@ -72,26 +68,37 @@ public struct FileAttachmentView: View {
...
         .sheet(isPresented: $fullScreenShown) {
             FileAttachmentPreview(attachment: attachment)
         }
         .accessibilityIdentifier("FileAttachmentView")
     }
+
+    private func retryUpload() {
+        let messageId = attachment.id.messageId
+        let cid = attachment.id.cid
+        let controller = chatClient.messageController(cid: cid, messageId: messageId)
```
martinmitrevski, line 99: hmm, is this ideal here? Shall we move it to the VM via the options?

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
```diff
@@ -267,9 +268,16 @@ public struct MessageMediaAttachmentsContainerView<Factory: ViewFactory>: View {
             height: height,
             cornerRadius: effectiveRadius,
             corners: effectiveCorners,
-            isOutgoing: message.isSentByCurrentUser
+            isOutgoing: message.isSentByCurrentUser,
+            onUploadRetry: item.uploadingState?.state == .uploadingFailed ? { [message, chatClient] in
+                guard let cid = message.cid else { return }
+                let controller = chatClient.messageController(
```
martinmitrevski, line 274: similarly here, we can have one method in the vm doing this

martinmitrevski (approved):
LGTM ✅  I would just move the resend logic to the vm

Sources/StreamChatSwiftUI/ChatMessageList/FileAttachmentView.swift
nuno-vieira: We don't have any view model here. I can set it up, but I'm not sure what additional changes will be required

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
nuno-vieira: Same here, there is no view model on this view, so we need to pass an existing one or create one for this.

I think we can do it later if we think it is worth it and customers request it

Sources/StreamChatSwiftUI/ChatMessageList/FileAttachmentView.swift
martinmitrevski: no, that's not what I meant. Just put it a handler onRetry or similar in the FileAttachmentViewOptions, and the factory method will pass the view model resend method. So just the method, not the whole VM - that will slow things down a lot.

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
martinmitrevski: anyway, if it's a lot of changes, ignore it

nuno-vieira: I've tried to change this, but TBH I prefer the current solution for now. Without a proper view model, it will just forward the action to the factory, which is not good. The factory should just create things, not do actions

PR #1428, opened 2026-04-20
Debounce SwiftUI Demo app user search to match UIKit demo app

testableapple:
### 🔗 Issue Links

Resolve https://linear.app/stream/issue/IOS-1262

### 🎯 Goal

Stop intermittent **“Error loading the users”** in the SwiftUI demo when searching for people to chat with / add to a group. Behavior should match the UIKit demo’s search pattern.

### 📝 Summary

- Debounce user search in **`NewChatViewModel`** and **`CreateGroupViewModel`** with **1000 ms** delay, matching **`CreateChatViewController`** / **`CreateGroupViewController`** in stream-chat-swift.
- Keep the **initial** `search(term: nil)` **immediate** on screen load (same as UIKit `viewDidLoad`).
- Add a **request generation** guard so completions from superseded searches don’t flip UI state.
- **`NewChatViewModel`**: after a successful fetch, use **`.noUsers`** when the result list is empty (aligned with UIKit empty state).

### 🛠 Implementation

SwiftUI was calling `ChatUserSearchController.search(term:)` on **every keystroke** via `searchText`’s `didSet`, while the UIKit demo batches changes with `DispatchWorkItem` and `asyncAfter(..., 1 second)`. That difference likely overloaded the backend (including repeated work for the default user-list query, which sorts by name in the SDK’s `UserListQuery.search(term:)`).

### 🧪 Manual Testing Notes

1. Run **DemoAppSwiftUI**, open **New Chat**.
2. Type quickly in the user search field: expect **spinner**, then results after ~1 s idle; no steady **“Error loading the users”** under normal backend load.
3. Open **Create a group** → **Add group members**, repeat search typing.
4. Confirm first load still shows users without an extra 1 s wait before the first request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Made user search in New Chat and Create Group more reliable by preventing stale results from replacing newer ones.
  * Added debounce to searches to reduce unnecessary requests and improve responsiveness while typing.
  * Improved loading and empty/error states so search feedback and initial user list behave more predictably.

### DIFF
--- a/DemoAppSwiftUI/ChannelHeader/NewChatViewModel.swift
+++ b/DemoAppSwiftUI/ChannelHeader/NewChatViewModel.swift
@@ -2,6 +2,7 @@
 // Copyright © 2026 Stream.io Inc. All rights reserved.
 //
 
+import Dispatch
 import StreamChat
 import StreamChatCommonUI
 import StreamChatSwiftUI
@@ -12,7 +13,7 @@ import SwiftUI
 
     @Published var searchText: String = "" {
         didSet {
-            searchUsers(with: searchText)
+            scheduleDebouncedUserSearch()
         }
     }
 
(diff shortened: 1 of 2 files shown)


martinmitrevski (approved):
might've been easier to use Combine's debounce, but this is also fine with me

DemoAppSwiftUI/ChannelHeader/NewChatViewModel.swift
```diff
@@ -49,11 +50,16 @@ import SwiftUI
     private lazy var searchController: ChatUserSearchController = chatClient.userSearchController()
     private let lastSeenDateFormatter = DateUtils.timeAgo
 
+    /// Matches UIKit demo `CreateChatViewController.throttleTime` / `CreateGroupViewController.throttleTime`.
+    private let userSearchDebounceMilliseconds = 1000
+    private var userSearchDebounceWorkItem: DispatchWorkItem?
+    private var userSearchRequestGeneration: UInt64 = 0
```
laevandus, line 56: One improvement could be checking if the search term changed or not before starting the search request. At the moment if I tap on the search field, it goes into loading state although nothing has changed (I did not type anything).

testableapple: A bit more code changes that i'd imagine, but i think i did what you requested. Please check out the last commit.

PR #1430, opened 2026-04-21
Fix Xcode 16.1 build

nuno-vieira:
### 🔗 Issue Links

`Build SDKs (Old Xcode)` (Xcode 16.1, Swift 6.0) and cron-only `Build Demo App (16.4, macos-15)` (Xcode 16.4, Swift 6.1) have been red on `develop` since #1425. Smoke/PR checks on Xcode 26.x stayed green, so the regression was only caught by the older-Xcode jobs.

### 🎯 Goal

Unblock the Xcode 16.1 / 16.4 builds without requiring the Swift 6.2 toolchain.

### 🔍 Root cause

The Xcode 16.1 failure was not a type-checking error — it was a **Swift 6.0 compiler crash** inside SILGen. The crash was hidden by `fastlane scan`'s formatter (xcbeautify/xcpretty), which swallowed the `swift-frontend` stack trace; only `** TEST BUILD FAILED ** (3 failures)` made it into the CI log. Re-running with the formatter disabled surfaced the real trace:

```
While evaluating request ASTLoweringRequest(
  Lowering AST to SIL for file ".../StreamChatSwiftUI/StreamChat.swift")
While silgen default arg initializer SIL function
  "@$s17StreamChatSwiftUI0aB0C10chatClient10appearance5utilsAcB0bF0C_...cfA1_"
 for expression at [StreamChat.swift:18:24 - line:18:30] RangeText="Utils("
Stack dump without symbol names ...
 0  swift-frontend  llvm::sys::PrintStackTrace
 4  swift-frontend  emitRawApply
 5  swift-frontend  swift::Lowering::SILGenFunction::emitApply
 ...
Command SwiftCompile failed with a nonzero exit code
```

That is an assertion in the SILGen lowering pass while synthesizing the **default-argument thunk** for

```swift
public init(
    chatClient: ChatClient,
    appearance: Appearance = Appearance(),
    utils: Utils = Utils()     // ← crashes here
) { ... }
```

After #1425, `Utils.init` gained a new default:

```swift
mediaLoader: MediaLoader? = nil
```

where `MediaLoader` is a `Sendable` protocol existential from `StreamChatCommonUI`. The combination of

- `@MainActor`-isolated `Utils`,
- a `nil` default for an optional protocol-existential parameter imported from another module, and
- `Utils()` itself being the default expression of **another** `@MainActor` initializer (`StreamChat.init`),

hits a SILGen bug that was silently fixed in Swift 6.2 (Xcode 26.x) by the closure/isolation inference and default-arg lowering rework. Because the compiler aborts before type-checking anything, no source-level change to the call sites could have fixed it — and that is why the previous attempt of pinning `@MainActor` on the `MediaLoader` completion closures had no effect.

### 🛠 Fix

Replace the crashing default on `Utils.init`

```swift
mediaLoader: MediaLoader? = nil
```

with a concrete, non-optional default that the SILGen thunk can lower without tripping the bug:

```swift
mediaLoader: MediaLoader = StreamMediaLoader(downloader: StreamImageDownloader())
```

and drop the now-unused `mediaLoader ?? …` fallback inside the init body. The public call-site signature of `Utils(...)` and `StreamChat(chatClient:)` is unchanged — existing callers still compile without edits.

### 📝 Summary

- `Sources/StreamChatSwiftUI/Utils.swift` — make `mediaLoader` non-optional with a concrete default.

### 🧪 Manual Testing Notes

N/A. CI should pass.

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `CDNRequester` is now passed in the constructor of `StreamMediaLoader` instead of `Utils` [#1425](https://github.com/GetStream/stream-chat-swiftui/pull/1425)
 
 ### 🐞 Fixed
+- Fix SDK not compiling with Xcode 16 [#1430](https://github.com/GetStream/stream-chat-swiftui/pull/1430)
 - Fix show/hide message translation animation [#1426](https://github.com/GetStream/stream-chat-swiftui/pull/1426)
 - Fix tapping a media attachment in the reactions overlay opening the fullscreen gallery [#1424](https://github.com/GetStream/stream-chat-swiftui/pull/1424)
 - Fix empty space around the previewed message in the reactions overlay not dismissing the overlay [#1424](https://github.com/GetStream/stream-chat-swiftui/pull/1424)
--- a/Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
@@ -1203,7 +1203,7 @@ final class FileAddedAsset {
 
         utils.mediaLoader.loadImage(
             url: imageAttachment.imageURL
-        ) { result in
+        ) { @MainActor result in
             if let image = (try? result.get())?.image {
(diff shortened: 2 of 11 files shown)


testableapple (approved):
🚀

PR #1431, opened 2026-04-22
Fix swipe-to-reply gesture conflicting with message list scrolling

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1644/v5-qa-swipe-to-reply-overrides-message-list-scrolling

### 🎯 Goal

The swipe-to-reply gesture on a message item was competing with the message list's vertical scroll gesture: vertical or diagonal drags sometimes got recognized as a swipe-to-reply, and the list occasionally refused to scroll, particularly on iOS 18+ where `ScrollView` gesture coordination is stricter.

This PR keeps the swipe-to-reply available across the whole message item view (not just the bubble) while making sure the scroll view always wins on any non-clearly-horizontal drag.

### 📝 Summary

- Switched `SwipeToReplyModifier` from `.gesture(DragGesture(...).updating($offset))` + `onChange(of: offset)` to `.simultaneousGesture(DragGesture(...).onChanged/onEnded)`.
- Bumped the default `MessageDisplayOptions.minimumSwipeGestureDistance` from `20` to `30`.

### 🛠 Implementation

- `.simultaneousGesture` lets the swipe-to-reply drag coexist with the `ScrollView`'s internal pan recognizer instead of competing with it, so scrolling is never blocked by the swipe gesture being attached.
- Raising the minimum swipe distance to `30` gives the scroll view a larger unambiguous window at the start of a drag before the swipe-to-reply can even recognize. In practice this means small vertical or diagonal movements now resolve to scroll, and only drags that are clearly intentional horizontal gestures trigger the reply.
- Replacing `@GestureState` + `onChange(of: offset)` with direct `onChanged` / `onEnded` handlers simplifies the state flow and removes a potential source of multi-per-frame state updates.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  n/a   |  n/a  |

### 🧪 Manual Testing Notes

- Scroll the list vertically and diagonally, both slowly and fast — scrolling should always feel responsive and never get "eaten" by the swipe gesture.
- While the list is idle, swipe a message horizontally (starting both on and outside the bubble, anywhere within the item view) — the reply indicator appears and, past the threshold, the message becomes the quoted reply.
- Verify on both iOS 18+ and earlier versions.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed swipe-to-reply gesture conflicting with message list scrolling
  * Refined swipe gesture sensitivity for improved reliability and responsiveness

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
+++ b/Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
@@ -257,9 +257,15 @@ struct SwipeToReplyModifier: ViewModifier {
 
     @State private var offsetX: CGFloat
     @State private var swipeExcludedFrames: [CGRect] = []
-    @GestureState private var offset: CGSize = .zero
 
     private let replyThreshold: CGFloat = 60
+
+    /// How much larger the horizontal drag component must be than the
+    /// vertical component before the drag is recognised as a swipe-to-reply.
+    /// Values > 1 bias ambiguous/diagonal drags toward the enclosing
+    /// scroll view, so the list keeps scrolling unless the user's motion is
+    /// clearly horizontal.
+    private let horizontalDominanceRatio: CGFloat = 2
 
     init(
         message: ChatMessage,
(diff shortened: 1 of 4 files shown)


Sources/StreamChatSwiftUI/ChatMessageList/MessageItemView.swift
```diff
@@ -374,17 +359,48 @@ struct SwipeToReplyModifier: ViewModifier {
...
+struct ConditionalSwipeToReplyModifier: ViewModifier {
+    let isEnabled: Bool
+    let message: ChatMessage
+    let channel: ChatChannel
+    let isSwipeToQuoteReplyPossible: Bool
+    @Binding var quotedMessage: ChatMessage?
+
+    @ViewBuilder
+    func body(content: Content) -> some View {
+        if isEnabled {
```
martinmitrevski: Conditional modifiers on a list item are not recommended for performance reasons, especially in this case, where it can frequently changed between these 2 states.

laevandus (approved):
I tried this branch and develop and felt like even develop worked fell enough. Swipe to reply works OK and did not bother me while just scrolling the message list. Therefore, OK to include this fine-tuning as well.

PR #1433, opened 2026-04-22
Fix voice recording gesture firing while mic button is hidden

nuno-vieira:
### 🔗 Issue Links

- [IOS-1646](https://linear.app/stream/issue/IOS-1646/v5-qa-voice-recording-gesture-is-active-when-command-ui-is-active)

### 🎯 Goal

The voice recording gesture overlay was active whenever the composer had no content, even when the mic button itself had been replaced by another trailing control (send button while an instant command like `/giphy` is active, confirm button while editing a message, or the slow-mode cooldown indicator). That kept the "hold to record" toast firing on taps the user could not associate with any visible control. The overlay was also active in frozen / no-send channels where the mic should never be available.

### 📝 Summary

- Introduce a single, shared resolver for `MessageComposerInputState` and have both `MessageComposerViewModel.composerInputState` and `ComposerInputView`'s local state call into it, so the trailing composer button and the voice recording gesture overlay derive from the **same source of truth**.
- Derive `shouldShowRecordingGestureOverlay` from that state: the overlay is only active while recording, or while the mic is the active trailing control (`composerInputState == .allowAudioRecording`) in a channel that allows sending.
- Gate `.allowAudioRecording` on `canSendMessage` so the mic (and its gesture overlay) no longer appear in frozen / no-send channels.
- Add unit tests for all the previously broken cases (instant command active, slow-mode cooldown, editing a message, no-send channel) and for the "recording in progress trumps everything" invariant.

### 🛠 Implementation

`VoiceRecordingGestureOverlay` in `MessageComposerView` is gated by `viewModel.shouldShowRecordingGestureOverlay`. Previously it returned `true` whenever `recordingState == .initial && !hasContent` (or while recording), which did not match the mic-button visibility rules encoded in `composerInputState` inside `ComposerInputView`. In particular, activating an instant command clears the composer text (so `hasContent == false`) and replaces the trailing button with the send button — yet the invisible gesture kept intercepting taps. The same `composerInputState` logic also ignored `canSendMessage`, so the mic was rendered in frozen channels too.

The fix extracts the decision tree into a shared `MessageComposerInputState` initializer and has both the view model and the view call into it:

```swift
extension MessageComposerInputState {
    init(
        cooldownDuration: Int,
        isEditingMessage: Bool,
        isInstantCommandActive: Bool,
        isVoiceRecordingEnabled: Bool,
        hasContent: Bool,
        canSendMessage: Bool
    ) {
        if cooldownDuration > 0 {
            self = .slowMode(cooldownDuration: cooldownDuration)
        } else if isEditingMessage {
            self = .editing(hasContent: hasContent)
        } else if isInstantCommandActive {
            self = .creating(hasContent: hasContent, hasCommand: true)
        } else if isVoiceRecordingEnabled && !hasContent && canSendMessage {
            self = .allowAudioRecording
        } else {
            self = .creating(hasContent: hasContent, hasCommand: false)
        }
    }
}
```

`MessageComposerViewModel` exposes the resolved state (and the derived overlay flag) for other consumers to read:

```swift
public var composerInputState: MessageComposerInputState {
    MessageComposerInputState(
        cooldownDuration: cooldownDuration,
        isEditingMessage: editedMessage?.wrappedValue != nil,
        isInstantCommandActive: composerCommand?.displayInfo?.isInstant == true,
        isVoiceRecordingEnabled: utils.composerConfig.isVoiceRecordingEnabled,
        hasContent: hasContent,
        canSendMessage: canSendMessage
    )
}

public var shouldShowRecordingGestureOverlay: Bool {
    if recordingState.isRecording { return true }
    guard recordingState == .initial else { return false }
    if case .allowAudioRecording = composerInputState, canSendMessage { return true }
    return false
}
```

`ComposerInputView`'s private `composerInputState` now calls the same initializer, so its trailing-button rendering and the view model's overlay gating cannot drift apart.

#### Why this approach (non-breaking)

An earlier iteration of this PR made `composerInputState` a required parameter on `ComposerInputView.init` and `ComposerInputViewOptions.init`, which would have been a breaking change for any integrator that overrides `ViewFactory.makeComposerInputView` and constructs `ComposerInputViewOptions` directly. To **avoid the breaking change**, the shared decision tree was moved down into a (non-public) `MessageComposerInputState` initializer that both call sites use:

- The view model exposes its computed `composerInputState` publicly for consumers that want to observe or reuse it.
- `ComposerInputView` keeps its existing public init unchanged and internally computes the same state locally via the shared initializer.

No public `init` signatures, options classes, or factory methods changed — integrators need to make no source-level adjustments — and we still avoid the duplicated decision tree that caused this bug.

### 🎨 Showcase

N/A – behavioural fix, no visual change when the mic is meant to be visible. The frozen-channel snapshot (`test_messageComposerView_frozenChannel.1.png`) is updated because the mic button is now correctly replaced by a disabled send button in that state.

### 🧪 Manual Testing Notes

1. Open a channel that supports commands in the demo app.
2. Type `/giphy` so the instant command chip appears and the trailing button becomes the send button.
3. Tap somewhere in the composer area — the "hold to record" toast should no longer appear.
4. Repeat while editing a message and while the channel has slow-mode cooldown active; no toast should appear in either case.
5. Open a frozen channel (or a channel without `sendMessage` capability) — the mic button should not be visible and tapping the composer area should not trigger the "hold to record" toast.
6. With an empty composer in a normal channel, long-press the mic and start a recording; the gesture overlay should still work normally (lock, cancel, release to send/save).

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed voice recording gesture and "hold to record" tip not displaying when the mic button is hidden.


martinmitrevski (commented):
We're not in beta mode any longer, we shouldn't be doing breaking changes.

nuno-vieira:
> We're not in beta mode any longer; we shouldn't be doing breaking changes.

I can avoid the breaking change, but it means we will have duplicated logic. Since we don't surface the view model to child views, it means whenever we introduce a new property in the view model, and it needs to be exposed to child views, and this new property can't have a default value, we will always have breaking changes. This one is very minor, so that is why I was considering introducing it. But I can revert it if we don't really want this simple breaking change

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
```diff
@@ -486,13 +486,40 @@ import SwiftUI
...
+            return .editing(hasContent: hasContent)
+        }
+        if composerCommand?.displayInfo?.isInstant == true {
+            return .creating(hasContent: hasContent, hasCommand: true)
+        }
+        if utils.composerConfig.isVoiceRecordingEnabled && !hasContent {
+            return .allowAudioRecording
+        }
+        return .creating(hasContent: hasContent, hasCommand: false)
+    }
```
laevandus, line 504: For avoiding breaking changes, MessageComposerInputState could have internal init method which takes all of these states to produce a value. Then it would be possible to init the enum in both view model and here without duplicating the logic. Not beautiful, but should get the work done.
But I get it, it is nicer to have it like we have it here, but compared to CDN change, I think it would be nice to not have more breaking changes.

laevandus (approved):
✅ Verified

martinmitrevski (approved):

PR #1434, opened 2026-04-23
Fix composer prompt views opening action sheet when returning from channel info

nuno-vieira:
### 🔗 Issue Links

N/A

### 🎯 Goal

In the `MessageComposerView`, the prompt views for camera, files and poll auto-presented their sheet/full-screen cover every time `.onAppear` fired. That's fine the first time the user taps the tab, but `.onAppear` also fires when the chat channel view reappears after a navigation pop (e.g. returning from the Channel Info screen), causing the sheet/cover to pop up again unexpectedly.

### 📝 Summary

- Replace `.onAppear` with the existing `.onLoad` modifier in `AttachmentFilePickerView`, `AttachmentCameraPickerView`, and `AttachmentPollPickerView` so the picker is only auto-presented on the first appearance.

### 🛠 Implementation

`OnLoadViewModifier` already guards its action with an `@State hasLoaded` flag, which is exactly the semantics these views want. Swapping `.onAppear` for `.onLoad` makes the auto-presentation fire once per view lifetime instead of on every re-appearance, preserving the original "user tapped the tab → open the picker" UX while preventing the re-present on navigation pop.

### 🎨 Showcase

N/A — behavioral fix, no visual change.

### 🧪 Manual Testing Notes

1. Open a channel.
2. Tap the attachment icon in the composer to expand the picker tray.
3. Switch to the Camera, Files, or Polls tab — the corresponding sheet/cover should present as before.
4. Dismiss the sheet.
5. Tap the channel header to open the Channel Info screen.
6. Tap Back to return to the channel.
7. The sheet/cover should **not** re-present. Previously, it would.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where attachment picker controls—including camera capture, file selection, and poll creation options—would incorrectly reappear after navigating back to a channel. These picker interfaces are now properly initialized and dismissed at the appropriate times, preventing unexpected re-presentation when returning to previously viewed conversations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
+- Fix attachment picker re-presenting after navigating back to the channel [#1434](https://github.com/GetStream/stream-chat-swiftui/pull/1434)
 
 # [5.1.0](https://github.com/GetStream/stream-chat-swiftui/releases/tag/5.1.0)
 _April 23, 2026_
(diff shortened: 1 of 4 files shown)


laevandus (approved):
✅  Verified manually and works

martinmitrevski (approved):

PR #1437, opened 2026-04-24
[CI] Improve impact analysis

testableapple:
### 🔗 Issue Links

Resolve https://linear.app/stream/issue/IOS-1556

### 🎯 Goal

Adopt a newer `fastlane-plugin-stream_actions` so `is_check_required` can distinguish **new PRs** (scan the full PR diff) from **new pushes** (scan only files changed in that `synchronize` push via `github.event.before` / `github.event.after`). That keeps smoke and related checks from re-running the whole PR file set on every commit when nothing under the watched paths changed in the **latest push**.

### 📝 Summary

- Bump `fastlane-plugin-stream_actions` in `fastlane/Pluginfile` and `Gemfile` / `Gemfile.lock` to the release that includes the `is_check_required` push-scoped behavior and `next_pr_number` action
- Rely on existing workflow `env` in `.github/workflows/smoke-checks.yml` (`GITHUB_EVENT_ACTION`, `GITHUB_EVENT_BEFORE`, `GITHUB_EVENT_AFTER`, `GITHUB_PR_NUM`) so Fastlane receives the event metadata without per-lane changes

### ⚠️ Important caveat

If you open a PR with some feature and then commit a change, for instance, in `CHANGELOG.md`, all checks will be skipped. For this particular case i implemented a new lane called `bundle exec fastlane pr` that prints the next available PR number. But i assume it is only one of many cases where this can happen.

### 🛠 Implementation

The plugin uses `gh pr view --json files` for `opened` (and other non-`synchronize` actions) and `gh api repos/.../compare/before...after` when `github.event.action == 'synchronize'` and valid shas and `GITHUB_REPOSITORY` are present. If compare is unavailable (e.g. some fork cases), it falls back to the full PR file list. No Fastfile change is required if `is_check_required` is already called with the usual `sources` and `github_pr_num` comes from `GITHUB_PR_NUM` as today.

### 🧪 Manual Testing Notes

1. Open a PR that only touches files **outside** the configured `sources` matrix paths → checks that use `is_check_required` should **skip** where appropriate.
2. Push a commit that only changes files under a given lane’s `sources` → that lane’s checks should run; pushes that don’t touch those paths should be **less likely to trigger** a full-PR re-evaluation for scopes that are push-aware.
3. Confirm `synchronize` runs on a normal PR (note: GitHub may not run `pull_request` workflows when the PR has **merge conflicts**).


laevandus (approved):
Let's test this

martinmitrevski (approved):

nuno-vieira (approved):

PR #1439, opened 2026-04-27
Fix image attachment flicker on reaction and Giphy not refreshing on shuffle

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1652

### 🎯 Goal

When a user added or removed a reaction on a message that contained image attachments, those images briefly showed the loading placeholder again — even though the same bitmap was already in Nuke's memory cache. The flicker happened because every reaction reflow re-signed the CDN URL with new `Policy`, `Signature`, and `Expires` query items, which broke the cache lookup that used `URL.absoluteString` as the key.

### 📝 Summary

- Stop the loading-placeholder flicker on cached image attachments after adding or removing a reaction.
- Refactor `StreamAsyncImage` into a single SwiftUI-driven view that owns its loading state and delegates caching entirely to `MediaLoader`/`StreamCDNRequester`.
- Allow callers to seed `StreamAsyncImage` with a pre-loaded `UIImage` (used by the reactions overlay) so the first frame never goes through the loading phase.
- Fix `AnimatedGifView` not updating when the Giphy attachment data changes (e.g. on shuffle).
- Remove the now-unused `NukeImageLoader` helper and its tests, and move the snapshot-test sync resolver onto `StreamAsyncImage` itself.

### 🛠 Implementation

- `StreamAsyncImage` now never re-enters the `.loading` phase once it has rendered an image. When the URL changes on a view that already shows an image, the previous bitmap stays on screen until the loader returns. For images already in Nuke's memory cache, that handoff happens within a single frame, so the user no longer sees a placeholder when reacting reflows the bubble with a freshly signed URL.
- The cache-key stability we rely on already lives in the LLC: `StreamCDNRequester.buildCachingKey` strips `Policy`, `Signature`, and `Expires` from CDN URLs, and `StreamImageDownloader` plumbs that key through Nuke's `ImageRequest.userInfo[.imageIdKey]`. We just stopped competing with it from the SwiftUI layer.
- A new `init(url:image:resize:content:)` lets callers (currently the reactions overlay, via `MessageContainerView` → `MessageRepliesView` etc.) seed an in-memory `UIImage`. When seeded, the view starts in `.success` and skips the loading phase entirely on first appearance.
- `AnimatedGifView` now implements `updateUIView`, re-applying the new `gifData` to the wrapped `UIImageView` whenever SwiftUI re-evaluates the representable. Without this, shuffling a Giphy kept the original animation on screen.
- `NukeImageLoader` and its tests are deleted; the snapshot-test sync resolver is now exposed via `StreamAsyncImageTestHooks.syncResolver` so tests still capture the loaded image instead of the empty placeholder.

#### Follow-up: fully avoiding the loading spinner

This PR removes the flicker for the common case (cached images), but the only way to **guarantee** zero loading state on the very first frame for arbitrary images is a synchronous cache lookup. SwiftUI's `.task` modifier always fires after the first render pass, so even a sub-millisecond cache hit can produce a one-frame placeholder.

To eliminate that completely we would need a synchronous `cachedImage(url:options:)` (or similar) on the `MediaLoader` protocol in the LLC, which `StreamAsyncImage` could call from `init` to seed `_phase` to `.success` before the first render. That is a follow-up because it changes the public `MediaLoader` protocol.

### 🧪 Manual Testing Notes

1. Send a message with an image attachment in a channel.
2. Add and remove a reaction on that message — the image should stay on screen the whole time, with no loading placeholder.
3. Long-press the message to open the reactions overlay — the attachment should appear instantly with no loading flicker.
4. Send a Giphy and tap "shuffle" — the displayed GIF should update each tap.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Render images from preloaded UIImage instances to seed initial display, with optional async URL-based updates; adds a convenience initializer for purely preloaded images.
  * Snapshot testing support for deterministic image rendering.

* **Bug Fixes**
  * Fixed image attachment flicker when reactions are added or removed.
  * Preserve previously displayed images on reload failures and avoid unnecessary transitions back to a loading state.


martinmitrevski (changes requested):
Are you sure about this one? It's still happening for me

Sources/StreamChatSwiftUI/CommonViews/StreamAsyncImage.swift
```diff
@@ -30,106 +40,165 @@ import SwiftUI
...
+    private var resolvedPhase: StreamAsyncImagePhase {
+        if case .success = phase { return phase }
+        if let url, let resolver = StreamAsyncImageTestHooks.syncResolver, let result = resolver(url, resize) {
+            return .success(result)
+        }
+        return phase
+    }
+
     private var taskIdentity: String {
         let urlPart = url?.absoluteString ?? ""
```
laevandus: Should we also use cache key for task identity because the full absolute url still contains query parameters which change and therefore the task runs more often while not updating the preview because cache hit is happening. Therefore, does unnecessary work.

Sources/StreamChatSwiftUI/CommonViews/StreamAsyncImage.swift
```diff
@@ -30,106 +40,165 @@ import SwiftUI
...
 
-        if case .success = initialPhase {
-            phase = initialPhase
-            return
+        // Only flip to the loading phase if no image has been rendered yet.
+        // When the URL changes on a view that already shows an image, we
+        // keep the previous image on screen until the loader returns the
+        // new one — for cached images that handoff is sub-frame, so the
+        // user never sees a placeholder.
+        if case .success = phase {} else {
```
laevandus: Yeah, this looks odd.

Sources/StreamChatSwiftUI/CommonViews/StreamAsyncImage.swift
martinmitrevski: that's a good point

Sources/StreamChatSwiftUI/CommonViews/StreamAsyncImage.swift
```diff
@@ -30,106 +40,165 @@ import SwiftUI
...
+        guard var resize else { return nil }
+        resize.width = resize.width.rounded()
+        resize.height = resize.height.rounded()
+        return resize
+    }
+}
+
+// MARK: - Test Hook
+
+/// Test-only hook used by snapshot tests to resolve image URLs
```
martinmitrevski: why do we have this in the prod code? It should be moved into the test target

Sources/StreamChatSwiftUI/CommonViews/StreamAsyncImage.swift
martinmitrevski: yeah, could be a bit nicer

Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(url: "https://github.com/GetStream/stream-chat-swift.git", from: "5.1.0")
+        .package(url: "https://github.com/GetStream/stream-chat-swift.git", revision: "46466a66843ea6d508fdd678ce657b191f0b96d4")
```
martinmitrevski: reminder for this

Sources/StreamChatSwiftUI/CommonViews/StreamAsyncImage.swift
nuno-vieira: Its not possible to be on the test target since it is not an extension. It is the only way to make it work in tests

Sources/StreamChatSwiftUI/CommonViews/StreamAsyncImage.swift
nuno-vieira: Yes I thought about this as well, but at this stage we don't have the caching key from the loader, so at most we can only duplicate the logic. But since I didn't see any improvements by changing it, I left it as it is

martinmitrevski (approved):
LGTM! ✅

PR #1451, opened 2026-05-12
Improve VoiceOver experience for the Create Poll sheet

nuno-vieira:
### 🔗 Issue Links

- [IOS-1691](https://linear.app/stream/issue/IOS-1691/swiftui-accessibility-attachment-picker-create-poll-sheet)

### 🎯 Goal

Address the Notion accessibility audit findings for the Create Poll sheet (Component = Attachment Picker / Create Poll). The audit numbers prefixed with `audit-` below refer to the rows in the audit table and are unrelated to GitHub issues.

### 📝 Summary

- `audit-34` — Post a screen change announcement naming the sheet on appear so VoiceOver users land with context.
- `audit-35` — Give the toolbar confirm button a real "Save Poll" label and drop the spurious `.isSelected` trait that the bordered-prominent style adds; the existing `.disabled()` already exposes the disabled trait correctly.
- `audit-36` — Mark the section spacer and the trailing list spacer as accessibility hidden so there is no empty focus stop between "Options" and "Multiple votes".
- `audit-37` — Make the option reorder handle a labelled adjustable element ("Reorder option, Option N of M") with an `accessibilityAdjustableAction` that calls a new view-model helper, so VoiceOver users can reorder via the rotor.
- `audit-38` — Label the Max Votes stepper buttons "Decrease vote limit" / "Increase vote limit", apply `.disabled(!enabled)` so the disabled trait reaches the bounds, and expose the current value via a "Vote limit, N" element.
- `audit-39` — Combine the title + description + switch rows in `CreatePollSettingCard` and the multiple-votes / vote-limit toggle rows into single accessibility elements via `.accessibilityElement(children: .combine)`, matching the focus order of the "Multiple votes" row.

### 🧪 Manual Testing Notes

1. Enable VoiceOver and open the channel composer.
2. Tap the attachment picker → Poll. VoiceOver should announce "Create Poll" on present and focus should land somewhere meaningful rather than on the bare close button.
3. Verify the trailing toolbar button announces as "Save Poll, button" (enabled) or "Save Poll, dimmed, button" (disabled before required fields are filled). It should not say "Selected" or "Dim".
4. Add a couple of options. Focus the drag handle on a non-empty option: VoiceOver should announce "Reorder option, Option N of M, adjustable". Swipe up / down to reorder, and confirm the order reflects in the list.
5. Enable "Multiple votes" and the "Limit votes per person" toggle. The stepper buttons should announce "Decrease vote limit" / "Increase vote limit", be dimmed at 2 (min) and 10 (max), and the static value should announce as "Vote limit, N".
6. Swipe through the Anonymous poll, Suggest an option and Add a comment rows. Each should announce as a single element with "{title}, {description}, switch, on/off" and activate the toggle when double-tapped.
7. Confirm there is no longer an empty focus stop between the Options section and the first settings row.

### ☑️ Contributor Checklist

## Release Notes

* **New Features**
  * Improved VoiceOver accessibility for Create Poll sheet with screen change announcements and option reordering through accessibility controls, including new accessibility labels for poll configuration and voting limits.

* **Documentation**
  * Updated changelog with accessibility enhancements for Create Poll.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-swiftui/pull/1451)

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatComposer/Polls/CreatePollView.swift
+++ b/Sources/StreamChatSwiftUI/ChatComposer/Polls/CreatePollView.swift
@@ -171,14 +171,10 @@ public struct CreatePollView<Factory: ViewFactory>: View {
     /// Posts a VoiceOver screen change announcement naming the sheet so users
     /// orient themselves on open instead of landing on the bare "Close" button.
     private func postScreenChangedAnnouncement() {
-        let title = L10n.Composer.Polls.createPoll
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-            if #available(iOS 17.0, *) {
-                AccessibilityNotification.ScreenChanged(title).post()
-            } else {
-                UIAccessibility.post(notification: .screenChanged, argument: title)
-            }
-        }
+        ComposerAccessibilityAnnouncer.announce(
+            L10n.Composer.Polls.createPoll,
+            kind: .screenChanged
+        )
     }
 
     @ViewBuilder
--- a/Sources/StreamChatSwiftUI/ChatComposer/Polls/CreatePollView.swift
+++ b/Sources/StreamChatSwiftUI/ChatComposer/Polls/CreatePollView.swift
@@ -45,6 +45,7 @@ public struct CreatePollView<Factory: ViewFactory>: View {
                     settingsSection
                     Spacer()
                         .modifier(CreatePollRowModifier(topSpacing: 0, bottomSpacing: 0))
+                        .accessibilityHidden(true)
                 }
                 .environment(\.defaultMinListRowHeight, 1)
                 .listStyle(.plain)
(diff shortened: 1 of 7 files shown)


Sources/StreamChatSwiftUI/ChatComposer/Polls/CreatePollView.swift
```diff
@@ -156,6 +165,20 @@ public struct CreatePollView<Factory: ViewFactory>: View {
         Color.clear
             .frame(height: tokens.spacingLg)
             .modifier(CreatePollRowModifier(topSpacing: 0, bottomSpacing: 0))
+            .accessibilityHidden(true)
+    }
+
+    /// Posts a VoiceOver screen change announcement naming the sheet so users
+    /// orient themselves on open instead of landing on the bare "Close" button.
+    private func postScreenChangedAnnouncement() {
```
martinmitrevski, line 173: We have this logic in a separate object in another PR, can we use that here as well?

nuno-vieira: Good call. PR #1450 merged in the meantime, so I've rebased onto develop and reused `ComposerAccessibilityAnnouncer`. Added a small `Kind` parameter (`.announcement` / `.screenChanged`) so the same helper covers both cases, since the sheet wants `.screenChanged` so VoiceOver refocuses on the new screen.

martinmitrevski (approved):

PR #1477, opened 2026-05-26
Add attachment bubble customization

laevandus:
### 🔗 Issue Links

[IOS-1728](https://linear.app/stream/issue/IOS-1728/swiftuibumble-attachment-bubble-customisation)

Docs PR: https://github.com/GetStream/docs-content/pull/1319

### 🎯 Goal

Allow customers to customise message attachment views, attachment item bubbles, and quoted-message reference bubbles without replacing Stream's default attachment views.

### 📝 Summary

- Add `Styles` hooks for message attachments views and per-attachment/reference item bubbles.
- Add `MessageAttachmentsViewModifierOptions` and `MessageAttachmentItemViewModifierOptions`.
- Extend `BubbleModifier` as the reusable primitive for custom attachment/reference bubble styling.
- Route file, voice recording, link preview, image, video, and chat quoted-message bubbles through the new style hook.
- Preserve default Giphy and poll rendering behavior.
- Add snapshot coverage for default and custom attachment bubble behavior.

### 🛠 Implementation

Example message attachments view customisation:

```swift
func makeMessageAttachmentsViewModifier(
    options: MessageAttachmentsViewModifierOptions
) -> some ViewModifier {
    MessageBubbleModifier(
        message: options.message,
        isFirst: options.isFirst,
        injectedBackgroundColor: UIColor.systemYellow,
        cornerRadius: 26
    )
}
```

Example attachment item / quoted-message bubble customisation:

```swift
func makeMessageAttachmentItemViewModifier(
    options: MessageAttachmentItemViewModifierOptions
) -> some ViewModifier {
    let showsBorder = options.attachmentType != nil
    return BubbleModifier(
        corners: .allCorners,
        backgroundColors: [.yellow.opacity(0.2)],
        borderColor: showsBorder ? .orange : nil,
        borderWidth: showsBorder ? 1 : 0,
        cornerRadius: 20,
        contentInsets: EdgeInsets(top: 6, leading: 6, bottom: 6, trailing: 6)
    )
}
```

`MessageAttachmentItemViewModifierOptions.attachmentType` is optional. `nil` is used for quoted-message reference bubbles, while file, voice recording, link preview, image, and video attachments provide their concrete attachment type.

### 🎨 Showcase

N/A

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added new modifier hooks for customizing attachment bubbles and individual attachment items.

* **Improvements**
  * Unified and simplified attachment styling for images, video, files, voice recordings, and link previews to ensure consistent rendering and layout.
  * Enhanced bubble border and padding configurability and refined media thumbnail clipping behavior.

* **Tests**
  * Added snapshot tests covering default vs. custom attachment bubble styles across attachment types.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-swiftui/pull/1477?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Sources/StreamChatSwiftUI/ChatMessageList/MessageAttachmentsView.swift
```diff
@@ -123,40 +163,66 @@ public struct MessageAttachmentsView<Factory: ViewFactory>: View {
...
+        } else {
+            content
         }
-        .accessibilityElement(children: .contain)
-        .accessibilityIdentifier("MessageAttachmentsView")
     }
 }
 
 /// Bubble styling configuration for stacked attachments.
 enum MessageAttachmentsBubbleConfiguration {
```
laevandus: I need to review some usage of this type since it modifies some styling related things based on if bubble should be shown or not.

Sources/StreamChatSwiftUI/ChatMessageList/MessageAttachmentsView.swift
```diff
@@ -123,40 +163,66 @@ public struct MessageAttachmentsView<Factory: ViewFactory>: View {
...
+                messageModifierInfo: messageModifierInfo,
+                bubbleInsets: bubbleInsets
+            )
+        )
+        .accessibilityElement(children: .contain)
+        .accessibilityIdentifier("MessageAttachmentsView")
+    }
+}
+
+private struct MessageAttachmentsBubbleModifier<Factory: ViewFactory>: ViewModifier {
```
martinmitrevski: why not expose a modifier directly via the factory like before?

nuno-vieira: Yes was thinking the same. Even if we don't do anything by default, we have this pattern where we use empty modifiers

StreamChatSwiftUITests/Tests/ChatChannel/__Snapshots__/MessageView_Tests/test_messageViewVoiceRecordingWithTextFromParticipantMultiple_snapshot.small-dark.png
nuno-vieira, line 1: This one we lost some padding in the middle of the attachments

Sources/StreamChatSwiftUI/ChatMessageList/QuotedMessageView/ChatQuotedMessageView.swift
```diff
@@ -41,16 +38,17 @@ public struct ChatQuotedMessageView<Factory: ViewFactory>: View {
...
                 quotedMessage: quotedMessage,
-                outgoing: parentMessageSentByCurrentUser
+                outgoing: parentMessage.isSentByCurrentUser
             )
         )
-        .modifier(ReferenceMessageViewBackgroundModifier(
-            backgroundColor: Color(
-                parentMessageSentByCurrentUser
-                    ? colors.chatBackgroundAttachmentOutgoing
-                    : colors.chatBackgroundAttachmentIncoming
```
nuno-vieira, line 51: Are we not using this anymore? Shouldn't it be used internally in our `MessageAttachmentItemViewModifier` implementation?

StreamChatSwiftUITests/Tests/ChatChannel/__Snapshots__/MessageView_Tests/test_messageViewVoiceRecordingWithTextFromParticipantMultiple_snapshot.small-dark.png
laevandus: This needs investigation. I think it is border color related.

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
```diff
@@ -0,0 +1,233 @@
...
+                    ),
+                    backgroundColor: defaultAttachmentBackgroundColor,
+                    borderColor: Color(colors.borderCoreDefault),
+                    cornerRadius: tokens.messageBubbleRadiusAttachment,
+                    corners: attachmentCorners(isSingleWithoutCaption: options.message.isSingleVoiceWithoutCaption)
+                ))
+            } else {
+                content
+            }
+        case .some(.linkPreview):
```
nuno-vieira, line 100: This name of the modifier is `DefaultMessageAttachmentItemViewModifier` but it uses `AttachmentContainerViewModifier`. Is this correct? I was expecting the same prefix, either `Item `or `Container.`

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
```diff
@@ -0,0 +1,233 @@
...
+
+    private func attachmentCorners(isSingleWithoutCaption: Bool) -> UIRectCorner {
+        options.isFirst && isSingleWithoutCaption
+            ? options.message.bubbleCorners(
+                isFirst: options.isFirst,
+                forceLeftToRight: utils.messageListConfig.messageListAlignment == .leftAligned,
+                layoutDirection: layoutDirection
+            )
+            : .allCorners
+    }
```
nuno-vieira, line 159: It could make sense to extract this into a view model or something to override this logic

Sources/StreamChatSwiftUI/ChatMessageList/AttachmentBubbleModifiers.swift
```diff
@@ -0,0 +1,233 @@
...
+        return attachmentCounts.count == 1 && (attachmentCounts[.image] == 1 || attachmentCounts[.video] == 1)
+    }
+
+    fileprivate var isSingleFileWithoutCaption: Bool {
+        text.isEmpty && fileAttachments.count == 1
+    }
+
+    fileprivate var isSingleVoiceWithoutCaption: Bool {
+        text.isEmpty && voiceRecordingAttachments.count == 1
+    }
```
nuno-vieira: Good that this is not public, but overall, this usually does not scale very well if we need more data outside of ChatMessage. Something we want to move out from: https://linear.app/stream/issue/IOS-1024/v5-remove-model-extensions-to-format-text-example-channelsubtitletext

Ideally, it would be nice for this to be part of `MessageItemViewModel` and let the modifiers use it, but it depends on how flexible we want to make this at the moment.

Sources/StreamChatSwiftUI/ChatMessageList/QuotedMessageView/ChatQuotedMessageView.swift
nuno-vieira: This is not exposed in the factory, but the goal was to eventually expose ReferenceMessageView in the view factory along with the ReferenceMessageViewModifier, so that customers could change the Reference styling (Edit + Quote) in one factory method

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
```diff
@@ -259,15 +257,13 @@ public struct MessageMediaAttachmentsContainerView<Factory: ViewFactory>: View {
...
-        let effectiveRadius = isSingleMediaWithoutCaption ? tokens.messageBubbleRadiusGroupBottom : cornerRadius
-        let effectiveCorners: UIRectCorner? = isSingleMediaWithoutCaption ? noCaptionBubbleCorners : nil
         return MessageMediaAttachmentContentView(
             factory: factory,
             source: item,
             width: width,
             height: height,
-            cornerRadius: effectiveRadius,
-            corners: effectiveCorners,
+            cornerRadius: 0,
```
martinmitrevski: why do we change these?

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
```diff
@@ -0,0 +1,233 @@
...
+    @Environment(\.layoutDirection) private var layoutDirection
+
+    let options: MessageAttachmentItemViewModifierOptions
+
+    init(options: MessageAttachmentItemViewModifierOptions) {
+        self.options = options
+    }
+
+    @ViewBuilder
+    func body(content: Content) -> some View {
```
martinmitrevski, line 72: are customers supposed to be doing all this by themselves when creating their own modifier?

martinmitrevski (commented):
Looks good in general, but I have few questions

Sources/StreamChatSwiftUI/ChatMessageList/QuotedMessageView/ChatQuotedMessageView.swift
laevandus: Not used anymore for chat, I replaced it with the new modifier because quotes are shown together with other attachments so it would make sense for quotes and attachments to share the same view modifier - single place to change attachment views.

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
laevandus: Yes, if they want complex logic like our new design has, then yes. Otherwise it is gonna be much shorter.

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
laevandus: I'll think about the name. AttachmentContainerViewModifier can be used for each attachment or for the whole stack.

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
laevandus: It is meant to be swapped out if anyone needs different styling rules for attachments. Our default design has already many rules and I think it is better to keep this internal. If we would want to add a view model, then it needs quite a bit of refactoring since the view model should not only capture these rules.

Sources/StreamChatSwiftUI/ChatMessageList/AttachmentBubbleModifiers.swift
laevandus: All of these are only used for attachments so I would prefer to keep them hidden here. I will think if these should be merged into the view modifier instead.

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
laevandus: Corner logic moved to `factory.styles.makeMessageAttachmentItemViewModifier`. I'll add a comment here. Otherwise it is not easy to override this because `effectiveRadius` and `effectiveCorners` are defined internally.

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
martinmitrevski: maybe we should consider exposing some default modifier which will handle some of this stuff (outside of the scope of this pr though)

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
martinmitrevski: I don't think it makes sense to have a view model in a view modifier. As mentioned above though, would be nice to expose some of the logic to customers at some point, it's a lot of boilerplate.

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentContentView.swift
```diff
@@ -79,12 +79,15 @@ public struct MessageMediaAttachmentContentView<Factory: ViewFactory>: View {
             thumbnail
         }
         .frame(width: width, height: height)
-        .clipShape(
-            BubbleBackgroundShape(
-                cornerRadius: cornerRadius,
-                corners: corners ?? [.topLeft, .topRight, .bottomLeft, .bottomRight]
-            )
-        )
+        .if(cornerRadius > 0, transform: { content in
```
martinmitrevski: we should avoid using if modifiers since they are not ideal for performance

Sources/StreamChatSwiftUI/ChatMessageList/MessageBubble.swift
```diff
@@ -87,32 +87,52 @@ public struct MessageBubbleModifier: ViewModifier {
...
+    ///   - cornerRadius: The bubble corner radius.
+    ///   - contentInsets: Insets applied to the content before drawing the bubble.
+    public init(
+        corners: UIRectCorner,
+        backgroundColors: [Color],
+        borderColor: Color? = nil,
+        borderWidth: CGFloat = 1,
+        cornerRadius: CGFloat = 18,
+        contentInsets: EdgeInsets = EdgeInsets()
+    ) {
```
laevandus, line 116: Removed the view modifier I added previously and reused BubbleModifier instead. Added missing arguments for supporting it:
- borderWidth
- contentInsets

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentContentView.swift
```diff
@@ -79,12 +79,15 @@ public struct MessageMediaAttachmentContentView<Factory: ViewFactory>: View {
...
+            content
+                .clipShape(
+                    BubbleBackgroundShape(
+                        cornerRadius: cornerRadius,
+                        corners: corners ?? [.topLeft, .topRight, .bottomLeft, .bottomRight]
+                    )
+                )
+        })
```
laevandus: Since this view is not doing the clipping anymore, I wrapped it with if. Clipping is done by the new attachments item view modifier.

Sources/StreamChatSwiftUI/ChatMessageList/DefaultMessageAttachmentsViewModifier.swift
laevandus: Ended up removing it and using BubbleModifier instead.

Sources/StreamChatSwiftUI/ChatMessageList/MessageBubble.swift
```diff
@@ -137,6 +149,20 @@ public struct BubbleModifier: ViewModifier {
             Color.clear
         }
     }
+
+    @ViewBuilder
+    private var borderOverlay: some View {
+        if borderWidth > 0 {
```
martinmitrevski, line 155: is an empty view returned here implicitly if borderWidth <= 0?

martinmitrevski (approved):
LGTM ✅

Sources/StreamChatSwiftUI/ChatMessageList/MessageBubble.swift
laevandus: Currently none

PR #1482, opened 2026-05-28
Add ChatChannelListItemViewModel and expose reusable sub-views

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1685/introduction-of-chatchannellistitemviewmodel
Docs: https://github.com/GetStream/docs-content/pull/1324

### 🎯 Goal

The channel list item view previously kept all of its presentation logic and sub-views inline as private properties. This made it hard for customers to:

- Customize individual pieces of the item (title, muted icon, preview variants) without reimplementing the whole channel item.
- Override the underlying presentation logic (timestamps, unread state, draft/typing/preview formatting, etc.).

This PR aligns `ChatChannelListItem` with the MVVM pattern already used by `ChatThreadListItem`, extracts the sub-views into a reusable, primitive-driven public API, and opens both view models for subclassing so customers can override individual presentation properties without forking the whole view.

### 📝 Summary

- Add `ChatChannelListItemViewModel`, an `open` class that owns all presentation logic for the channel list item. Properties with presentation logic are `open` for overriding; properties that read straight from the domain models are `public`.
- Extract reusable, public sub-views, all driven by primitive values so they can be composed into custom layouts:
  - `ChannelItemTitleView` — channel name plus the optional inline muted icon.
  - `ChannelItemMutedIcon` — the muted glyph (with a localized accessibility label).
  - `ChannelItemPreviewView` — the second line that renders the channel's latest activity.
  - Per-variant preview views: `ChannelItemFailedToSendView`, `ChannelItemDraftPreviewView`, `ChannelItemDeletedPreviewView`, and `ChannelItemMessagePreviewView` for the regular-message variant.
  - `ChannelItemAttachmentIcon` — the standard attachment glyph.
- Introduce `ChannelItemPreview`, a struct with static factory methods that behaves like an enum but stays source-compatible if new variants are added later. The regular-message variant carries a nested `MessageContent` value that combines an optional author prefix and an optional attachment icon. The typing variant carries the `ChatChannel` so `ChannelItemPreviewView` can render it through `ViewFactory.makeSubtitleTypingIndicatorView(options:)`, honoring customer overrides.
- Make `ChatChannelNavigatableListItem` generic over the channel item and inject it directly (mirroring `ChatThreadListNavigatableItem`), so the navigation wrapper can be reused around any custom channel item. The default `ChatChannelListItem` is now built by the factory.
- Extract `ChatThreadListItemViewModel` into its own file and expose `thread` as `public let`.
- Allow overriding the channel name via the `channelName` `open var` while keeping the value provided to the initializer as the default.
- Expose a localized accessibility label on `ChannelItemMutedIcon` so VoiceOver announces the channel's muted state.
- Name the channel item's second-line API "preview" so the naming reflects what the slot actually displays (latest message, typing, draft, deleted, or failed-to-send).

### 🛠 Implementation

- **View models** — `ChatChannelListItemViewModel` owns the channel item's presentation logic; `ChatThreadListItemViewModel` is extracted into its own file and opened for subclassing. Helper values that only support internal branching stay `private`; values consumed by the views are exposed (`open` when they carry presentation logic, `public` when they are straight reads of the domain models).
- **Preview value type** — `ChannelItemPreview` replaces ad-hoc branching with a value type produced by the view model. It is a `struct` with a private `Kind` enum and `public static` factory methods (`failedToSend()`, `typing(channel:)`, `draft(text:)`, `deleted(isSentByCurrentUser:)`, `message(_:)`), so new variants can be added later without source-breaking changes. The regular-message case carries a nested `ChannelItemPreview.MessageContent` (preview text plus optional author prefix and attachment icon).
- **Typing variant through the factory** — `ChannelItemPreviewView` is generic over `Factory` and renders the typing variant via `factory.makeSubtitleTypingIndicatorView(options:)`. Its primary init defaults the factory to `DefaultViewFactory.shared`; `ChatChannelListItem` passes its own factory through so customer typing-indicator overrides keep working.
- **Composable navigation wrapper** — `ChatChannelNavigatableListItem` becomes `ChatChannelNavigatableListItem<ChannelListItem, ChannelDestination>` and accepts the channel item directly, exactly like `ChatThreadListNavigatableItem`. `DefaultViewFactory.makeChannelListItem` now builds the default `ChatChannelListItem` and injects it.
- **Reusable sub-views** — the title, muted icon, attachment icon, and per-variant preview views are public and primitive-driven so customers can re-lay-out the channel item while still using the default view model to populate the data.

### ♻️ Backwards compatibility

- `ChatChannelListItem`'s existing `init` is preserved; a new convenience init that takes a `ChatChannelListItemViewModel` is added.
- All previously private/internal preview helpers are now public — additive, not source-breaking.
- The previous `ChatChannelNavigatableListItem` initializer that builds the default item from a `factory`/`channelName` is kept as a `@available(*, deprecated)` forwarding overload, with a message pointing to the new injection-based initializer.
- `ChannelItemPreview` is a `struct` with `static` factories rather than an `enum`, so new variants can be added later without breaking customers.

### 🧪 Manual Testing Notes

- Open `DemoAppSwiftUI`, browse the channel list, and verify titles, timestamps, unread badges, muted icons, draft previews, typing indicators, failed-to-send markers, deleted previews, and attachment previews all render as before.
- Open the thread list view and verify parent message text, latest reply, draft replies, unread counts, and author rows are unchanged.
- Run VoiceOver over a muted channel item and confirm it announces the muted state.
- Optional: subclass `ChatChannelListItemViewModel` in the demo app, override `channelName` or `messagePreviewText`, and confirm the item picks up the override.

### ☑️ Contributor Checklist

- [x] Documentation has been updated in the `docs-content` repo

## Summary by CodeRabbit

* **Refactor**
  * Channel and thread list items now use dedicated view models and smaller reusable subviews for clearer title/timestamp, unread badges, read indicators, muted icons, and preview/attachment rendering.
  * Preview rendering consolidated into a preview model with correct precedence for failed-to-send, typing, drafts, deleted, and regular messages (with optional author prefix and attachment icon).

* **New Features**
  * Channel navigation wrapper accepts a custom channel item view.

* **Tests**
  * Added tests covering channel item presentation and preview precedence.

* **Documentation**
  * Changelog updated to note new view models and reusable subviews.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-swiftui/pull/1482?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
```diff
@@ -28,291 +23,356 @@ public struct ChatChannelListItem<Factory: ViewFactory>: View {
...
+    @Injected(\.tokens) private var tokens
+
+    public init() {}
+
+    public var body: some View {
+        HStack(spacing: tokens.spacingXxs) {
+            Image(uiImage: images.messageListErrorIndicator)
                 .customizable()
-                .frame(width: tokens.iconSizeSm, height: tokens.iconSizeSm)
+                .frame(width: 16, height: 16)
```
laevandus: Do we wanna add these iconSizeSm back or 16 is what we want?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
```diff
@@ -0,0 +1,240 @@
...
+    open var readUsers: [ChatUser] {
+        channel.readUsers(
+            currentUserId: chatClient.currentUserId,
+            message: previewMessage
+        )
+    }
+
+    /// A boolean value indicating whether the read indicator should
+    /// show the delivered state.
+    open var showDelivered: Bool {
```
laevandus: Others use `shouldShow`

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
```diff
@@ -0,0 +1,240 @@
...
+
+import StreamChat
+import SwiftUI
+
+/// The view model for the channel list item view.
+///
+/// It contains the default presentation logic for the channel list item data.
+/// Subclass and override the `open` properties to customize what the channel
+/// list item displays.
+@MainActor open class ChatChannelListItemViewModel {
```
laevandus, line 13: Currently everything is open, but I am not sure if it is a good thing when it comes to maintaining it later. For example, config wrapping properties could just be internal bc one should update config instead of overriding it here. I think I would keep all the simple wrapper properties as internal for now. WDYT

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
nuno-vieira: This is probably a mistake, I will check, thanks!

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
nuno-vieira: yup makes sense to be consistent

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
nuno-vieira: I think it could be useful if customers want to override the behaviour in some scenarios. Where in a condition it should respect the dashboard, but in another condition it shouldn't. I feel like having customers always asking to open things is not a really good experience for them IMO

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemPreviewView.swift
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    let kind: Kind
+
+    private init(_ kind: Kind) {
+        self.kind = kind
+    }
+
+    /// Failed-to-send variant: shown when the last message failed to send.
+    public static func failedToSend() -> ChannelItemPreview {
```
martinmitrevski: why are all these methods?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
martinmitrevski: I agree with Toomas, and I think that some of those shouldn't be exposed. For example, why would you customize `unreadCount`, `hasUnread` or `readUsers`? Any other implementation you provide would be a fake one - since this is just pure data. If you, for example, change `readUsers`, then you allow this data to be manipulated, and this is sensitive for workplace customers.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /// A boolean value indicating whether the read indicator should
+    /// show the delivered state.
+    open var showDelivered: Bool {
+        previewMessage?.deliveryStatus(for: channel) == .delivered
+    }
+
+    /// The local message state of the preview message.
+    open var previewMessageLocalState: LocalMessageState? {
```
martinmitrevski: why is this exposed?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
```diff
@@ -0,0 +1,213 @@
...
+
+    // MARK: - Preview
+
+    /// The preview variant to render for the channel list item.
+    ///
+    /// Pass it to ``ChannelItemPreviewView``. The default
+    /// implementation picks the first applicable variant in this order of
+    /// precedence: failed-to-send, typing, draft, deleted, then a regular
+    /// message.
+    open var preview: ChannelItemPreview {
```
martinmitrevski: I think if we go to this level - better to build your own view? We expose all this complex VM stuff, where you have to learn all these details, but if you need to change this stuff - you're better off building your own views.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemPreviewView.swift
```diff
@@ -0,0 +1,153 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import SwiftUI
+
+/// Describes which variant the channel list item preview should render.
```
martinmitrevski: Most of these AI generated docs are noise - when you read it, it doesn't make sense. For example:
```
/// Built with the provided static factory methods. The underlying representation
/// is intentionally hidden so new variants can be added in the future without
/// breaking source compatibility for clients that switch on it.
```
How is this helpful for customers?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
```diff
@@ -5,49 +5,92 @@
...
+    /// mutate the `selectedChannel` binding (typically by invoking the
+    /// `onItemTap` callback from ``ChannelListItemOptions``) so the
+    /// navigation overlay can push the configured destination.
+    ///
+    /// - Parameters:
+    ///   - channelListItem: The custom view rendered as the row content.
+    ///   - factory: View factory used by sibling wrappers (kept for
+    ///     generic-parameter inference; not used to build the row).
+    ///   - channel: The channel represented by this row.
+    ///   - handleTabBarVisibility: Whether to hide the tab bar when the
```
martinmitrevski: why only 16.0 to 16.2?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
```diff
@@ -5,49 +5,92 @@
...
+///
+/// Use the convenience initializer (`init(factory:channel:channelName:...)`)
+/// to wrap the default ``ChatChannelListItem`` with navigation, or the
+/// composable initializer (`init(channelListItem:factory:channel:...)`) to
+/// inject a custom row view between this wrapper and any outer wrapper such
+/// as ``ChatChannelSwipeableListItem``.
+public struct ChatChannelNavigatableListItem<
+    Factory: ViewFactory,
+    ChannelDestination: View,
+    ChannelListItem: View
```
martinmitrevski: I don't see much point in generalizing this item, when it's just a navigation link

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
```diff
@@ -5,49 +5,92 @@
...
+        handleTabBarVisibility: Bool = true,
+        selectedChannel: Binding<ChannelSelectionInfo?>,
+        channelDestination: ((ChannelSelectionInfo) -> ChannelDestination)? = nil
+    ) {
+        self.channel = channel
+        self.handleTabBarVisibility = handleTabBarVisibility
+        _selectedChannel = selectedChannel
+        self.channelDestination = channelDestination
+        content = { _ in channelListItem }
+        _ = factory
```
martinmitrevski: why do we do this?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
```diff
@@ -5,49 +5,92 @@
...
+    }
+
+    /// Convenience initializer that constructs the default
+    /// ``ChatChannelListItem`` internally.
+    public init(
+        factory: Factory,
```
martinmitrevski: this is a breaking change without a default value.

Sources/StreamChatSwiftUI/ChatThreadList/ChatThreadListItemViewModel.swift
```diff
@@ -0,0 +1,161 @@
...
+
+import StreamChat
+import SwiftUI
+
+/// The view model for the thread list item view.
+///
+/// It contains the default presentation logic for the thread list item data.
+/// Subclass and override the `open` properties to customize what the thread
+/// list item displays.
+@MainActor open class ChatThreadListItemViewModel {
```
martinmitrevski, line 13: Has something changed here, or is it just moved to a new file?

Sources/StreamChatSwiftUI/CommonViews/TypingIndicatorView.swift
```diff
@@ -106,18 +106,35 @@ public struct TypingIndicatorDotsView: View {
...
+    /// Derives the typing text from the channel's currently-typing users.
     public init(channel: ChatChannel) {
         self.channel = channel
+        text = nil
+    }
+
+    /// Renders the provided typing text directly. Use this when the text is
+    /// already computed (e.g. by a view model) and overrides what would
+    /// otherwise come from `ChatChannel.typingIndicatorString(currentUserId:)`.
+    public init(text: String) {
```
martinmitrevski: Feels weird to do this - isn't it easier to just expose a way to overwrite the default method implementation?

martinmitrevski (commented):
I don't like this PR tbh - too much complexity, and I think we expose too many things. It's unclear where we draw the line between building your own view and using all these configs. Also, some unnecessary generalizations and too much AI noise.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
nuno-vieira: This has quite a lot of logic around tab bars etc... If customers want to keep the same logic we have, they need to fork all of this.

When I'm a customer, and I want to customise the channel list item view only, I would expect to only provide this view, not a Swipeable and Navigatable wrapper. Thread List works like this: you just need to provide the item view. But here, since we also need to provide the wrappers, I think at least we need to make it easy for customers to reuse these wrappers. Customers are from multiple levels of expertise and not all of them will understand that they will need to create a Navigation Link etc...

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
nuno-vieira: This is a new init

Sources/StreamChatSwiftUI/ChatThreadList/ChatThreadListItemViewModel.swift
nuno-vieira: Just moved to a new file and made it open

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
nuno-vieira: The comment was generated because of this line:

```swift
channelDestination(channel.channelSelectionInfo)
                            .modifier(
                                HideTabBarModifierForiOS16(
                                    handleTabBarVisibility: handleTabBarVisibility
                                )
                            )
```

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
nuno-vieira: This was still some left overs, the PR was still in Draft, so some stuff still needed some cleanup

Sources/StreamChatSwiftUI/CommonViews/TypingIndicatorView.swift
nuno-vieira: This was still a draft, I reverted back to how it was because changing it would make the code way too complex. It would be nice to be able to override this from the view model also, but at the moment it is good enough

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemPreviewView.swift
nuno-vieira: Yes this is implementation detail, will remove it

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
nuno-vieira: It was still in draft, this is now public only

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemPreviewView.swift
nuno-vieira: These methods are to simulate an enum, without actually exposing an enum to customers, to avoid future breaking changes. 

The channel preview is quite complex; it has a lot of logic. We have customers who want to reuse the channel preview logic and view from us, without rebuilding it from scratch, so this is a nice way for customers to be able to use the `ChannelItemPreviewView(preview: viewModel.preview)`, and they will have the preview for free on their custom channel list item view.

At the moment, we have the following supported previews:
- `.failedToSend()`
- `.typing()`
- `.draft()`
- `.deleted()`
- `.message()`

But in the future we can have also reactions etc, so we can add more cases without breaking changes

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
nuno-vieira: Fixed it, all properties that fetch data directly from domain models are now public

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
nuno-vieira: The ChatChannelListItem is not a simple view; it has a lot of logic in place. Some customers want to get this logic for free, and not repeat it from scratch, as well as have these views for free, without rebuilding them from scratch.

The Channel Preview is actually the most complex on here, and it is the one that we should provide to customers as a built-in component. They can decide whatever layout they want in the Channel List Item, and by just providing: `ChannelItemPreviewView(preview: viewModel.preview)` they will get all of our channel preview logic.

nuno-vieira:
> I don't like this PR tbh - too much complexity, and I think we expose too many things. It's unclear where we draw the line between building your own view and using all these configs. Also, some unnecessary generalizations and too much AI noise.

If you mean complexity on our side, maybe, but it's really just splitting things, not really a big deal. Even if it brings complexity on our side, the good thing is that it makes things easier for customers to reuse our logic and UI. 

Docs here: https://github.com/GetStream/docs-content/pull/1324

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItem.swift
nuno-vieira: This was already like this, thats why AI copied it. But I've changed it to use the tokens.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
martinmitrevski: it's fine to expose pre-built components and let customers re-use them when building their own views. But exposing a config like this is pushing the SDK towards subclassing over composition.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemPreviewView.swift
martinmitrevski: still don't see the benefit of this being a method - can't it be a property instead?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
martinmitrevski: sorry, I still don't get it - we're talking about few lines of code in total, better to just expose the modifier if needed

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
martinmitrevski: ok, indeed it seems accurate

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
martinmitrevski: then where's the old one?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
```diff
@@ -68,9 +74,40 @@ public struct ChatChannelNavigatableListItem<Factory: ViewFactory, ChannelDestin
...
         }
     }
+}
 
-    private var isSelected: Bool {
-        selectedChannel?.channel.cid == channel.cid
+public extension ChatChannelNavigatableListItem {
+    /// Convenience initializer that builds the default ``ChatChannelListItem``
+    /// internally and wraps it with navigation behavior.
+    @available(
```
martinmitrevski: Feels unnecessary to deprecate this. Plus, it's still a breaking change - there's no default param for the view factory.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemPreviewView.swift
nuno-vieira: It can't be a property, because you need to pass data, and the data is different per type. So it's like having an enum with arguments

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelNavigatableListItem.swift
nuno-vieira: Its deprecated now

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListItemViewModel.swift
```diff
@@ -0,0 +1,219 @@
...
+
+    // MARK: - Title row
+
+    /// The display name of the channel.
+    ///
+    /// The default implementation returns the channel name passed to the
+    /// initializer. Subclasses can override this to provide a custom display
+    /// name (for example, derived from custom channel data) and may call
+    /// `super.channelName` to fall back to the initializer value.
+    open var channelName: String { providedChannelName }
```
martinmitrevski, line 33: this one is debatable whether it should be open - but I'm leaving it up to you

martinmitrevski (approved):
lgtm ✅

PR #1496, opened 2026-06-17
Collapse composer picker only when multiple items are available

laevandus:
### 🔗 Issue Links

Resolves: [IOS-1767](https://linear.app/stream/issue/IOS-1767)

### 🎯 Goal

Do not collapse composer leading items when there is just one item (e.g. commands are disabled)

### 📝 Summary

- Collapse the composer's attachment picker on text entry **only when more than one picker item is available** (media + instant commands). A single-button picker now stays expanded while typing.
- Add `availablePickerItemsCount`, `fileUploadsAvailable`, and `configuredCommandsAvailable` on `MessageComposerViewModel`; reuse the latter two in the picker view and `showCommandsOverlay`.

### 🛠 Implementation

- The collapse-on-type lived in `MessageComposerViewModel.text`'s `didSet` and fired on every keystroke. It's now gated on `availablePickerItemsCount > 1`. The `else` branch (active slash-command or large paste) is unchanged.
- `availablePickerItemsCount` mirrors the button-visibility conditions in `AttachmentPickerTypeView`'s `.expanded` case: media via `canUploadFile`, commands via `config.commands.count`, both gated by `isSendMessageEnabled`.
- Extracted `fileUploadsAvailable` / `configuredCommandsAvailable` to remove duplication; the picker view's media button and `showCommandsOverlay` now route through them (same source, no behavior change).
- Note: `develop` does not need this fix — the composer was rewritten there and no longer collapses the picker on typing.

### 🧪 Manual Testing Notes

- **Single item** (uploads enabled, no commands configured, or vice-versa): start typing — the paperclip stays put, no swap to the arrow.
- **Multiple items** (uploads enabled *and* commands configured): start typing — the picker collapses to the arrow as before; tapping re-expands to both buttons.
- Selecting a slash command or pasting a large block still collapses (unchanged `else` branch).

### ☑️ Contributor Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### 🐞 Fixed
+- Fix the composer attachment picker collapsing the paperclip icon into an arrow while typing when only a single picker item is available [#1496](https://github.com/GetStream/stream-chat-swiftui/pull/1496)
+
 ### 🔄 Changed
 
 # [4.101.1](https://github.com/GetStream/stream-chat-swiftui/releases/tag/4.101.1)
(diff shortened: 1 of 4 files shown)


martinmitrevski (approved):
lgtm ✅

PR #1501, opened 2026-06-23
Improve scrolling performance of the channel list

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1746/swiftui-channel-list-scrolling-feels-laggish.

### 🎯 Goal

Channel-list scrolling was janky. Time Profiler showed the dominant cost was on-CPU avatar image resizing (ImageProcessingExtensions.byResizing → UIImage.draw), repeated on every cache miss, plus repeated PersonNameComponentsFormatter work for avatar initials. This reduces that work so scrolling stays smooth.

### 📝 Summary

- Persist resized avatar/image thumbnails to an on-disk cache so they are no longer re-decoded and re-resized on the CPU after the in-memory cache evicts them.
- Widen the image processing and decompression concurrency so bursts of avatar loads during fast scrolling use more cores.
- Cache computed avatar initials to avoid repeated locale-aware name parsing on the main thread.

### 🛠 Implementation

- Image pipeline (StreamImageDownloader): the stock Nuke pipeline (.withURLCache) only persists original downloaded data to an HTTP URLCache; processed images live only in the cost-limited in-memory cache. While scrolling a long list the memory cache evicts resized avatars, so every reappearance re-runs byResizing/UIImage.draw. The downloader now uses a dedicated pipeline built from .withDataCache with dataCachePolicy = .automatic, which stores the processed (resized) image for any request carrying a resize processor. Subsequent loads decode the already-resized thumbnail straight from disk. A dedicated pipeline (rather than mutating ImagePipeline.shared) avoids overriding any pipeline a host app configured. Disk keys stay stable across URL re-signing because requests already key on the CDN caching id.
- Queue concurrency: imageProcessingQueue and imageDecompressingQueue raised from 2 to 4.
- Initials (UserAvatar.initials(from:)): results are memoized in an NSCache keyed by name. The same names recur across many cells, and PersonNameComponentsFormatter is comparatively expensive and runs on the main thread.

### 🎨 Showcase

The hangs caused by the avatars and initials are now gone. We can still do some LLC optimizations.

Screenshot 2026-06-23 at 13 34 05

### 🧪 Manual Testing Notes

Scroll a channel list with many avatars (especially after the in-memory cache fills). Fling up and down repeatedly; previously-seen avatars should reappear without re-resizing. Gains are largest on scroll-back and on warm app launches (first-ever load of each avatar still pays one decode+resize). Unit tests cover the pipeline configuration and initials caching.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Performance Improvements**
  * Avatar initials are now cached to improve display performance when rendering user avatars, reducing unnecessary recomputation across the application.
  * Image downloading pipeline has been optimized with enhanced resource management and concurrent operations for faster and more efficient image loading throughout the interface.


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add support for enhanced mention suggestions [#1497](https://github.com/GetStream/stream-chat-swiftui/pull/1497)
 - Add `PollsConfig.maxQuestionLength` and `PollsConfig.maxOptionLength` to limit the character count of poll questions and options [#1500](https://github.com/GetStream/stream-chat-swiftui/pull/1500)
 
+### 🔄 Changed
+- Improve channel list scrolling performance [#1501](https://github.com/GetStream/stream-chat-swiftui/pull/1501)
```
nuno-vieira: Usually, for performance fixes, we have this section:
```suggestion
### ⚡️ Performance
- Improve channel avatar loading performance in the channel list [#1501](https://github.com/GetStream/stream-chat-swiftui/pull/1501)
```
Also, good to be more specific with the changelog, otherwise we will have a lot of "Channel list performance fix" in multiple changelogs, which will look the same.

"**Changed**" section should be used for breaking or behaviour changes.

nuno-vieira (approved):
Codewise LGTM ✅ Just added improvement to the Changelog 👍

CHANGELOG.md
martinmitrevski: makes sense, updated

PR #1502, opened 2026-06-25
Fix composer accessibility issues (VoiceOver focus, Dynamic Type)

nuno-vieira:
### 🔗 Issue Links
- https://linear.app/stream/issue/IOS-1776/a11y-message-view-does-not-focus-the-composer-on-entry
- https://linear.app/stream/issue/IOS-1785/a11y-composer-placeholder-and-also-send-in-channel-truncate-at-large
- https://linear.app/stream/issue/IOS-1784/a11y-snackbar-truncates-at-larger-text-sizes-instead-of-wrapping
- https://linear.app/stream/issue/IOS-1783/a11y-message-composer-text-does-not-scale-with-dynamic-text-size
- https://linear.app/stream/issue/IOS-1789/a11y-composer-placeholder-and-input-are-two-voiceover-focus-stops

### 🎯 Goal

Improve accessibility in the composer area and snackbar to comply with WCAG criteria for focus order (2.4.3) and text resize (1.4.4/1.4.10).

### 📝 Summary

- VoiceOver focus now lands on the composer text input when entering a channel, instead of the navigation bar back button
- Composer placeholder text wraps and scales with Dynamic Type instead of truncating
- "Also send in channel" label wraps at large text sizes instead of being clipped
- Snackbar text wraps at large text sizes instead of truncating on a single line

### 🛠 Implementation

- **VoiceOver initial focus (IOS-1776):** Added `didMoveToWindow()` override in `InputTextView` that posts a `UIAccessibility.screenChanged` notification targeting the text input after a short delay when VoiceOver is running. A flag prevents repeated focus hijacking on subsequent layout passes. The placeholder label is also marked as non-accessible to avoid a redundant focus stop when swiping.
- **Composer placeholder resize (IOS-1785):** Configured the placeholder `UILabel` with `numberOfLines = 0` and `adjustsFontForContentSizeCategory = true` so it wraps and scales with Dynamic Type. Added a `placeholderFittingHeight` computed property and a `onPlaceholderHeightChanged` callback to drive the SwiftUI height binding from actual placeholder layout. The `textFieldHeight` in `MessageComposerView` now only applies the max height cap when text is present, allowing the placeholder to expand freely.
- **Send in channel label (IOS-1785):** Removed `.lineLimit(1)` from the label in `SendInChannelView`.
- **Snackbar wrapping (IOS-1784):** Removed `.lineLimit(1)` from `SnackBarView` and added `.fixedSize(horizontal: false, vertical: true)` so text wraps inside the overlay.

### 🎨 Showcase

_N/A — accessibility improvements verified with VoiceOver and Dynamic Type on device._

### 🧪 Manual Testing Notes

1. **VoiceOver focus:** Enable VoiceOver → Open a channel from the channel list → VoiceOver should announce the composer text input, not the back button.
2. **Placeholder wrapping:** Set Dynamic Type to an accessibility size → Open a frozen channel (or one without send permission) → The placeholder text should wrap fully without truncation.
3. **Send in channel:** Set Dynamic Type to a large size → Open a thread → The "Also send in channel" label should wrap instead of being cut off.
4. **Snackbar wrapping:** Set Dynamic Type to an accessibility size → Tap and release the voice recording button → The snackbar text should be fully visible and wrapped.

### ☑️ Contributor Checklist

## Summary

* **Bug Fixes**
  * Improved composer input height syncing so placeholder sizing remains correct at large accessibility Dynamic Type sizes.
  * Updated composer height clamping so the maximum height applies only when there’s entered text.
  * Enabled wrapping for “send in channel” and snack bar text at larger text sizes.
  * Fixed VoiceOver focus behavior when entering a channel (now triggered once).

* **Tests**
  * Added snapshot coverage for extra-extra-large and accessibility text-size configurations.
  * Added `InputTextView` placeholder sizing tests.


Sources/StreamChatSwiftUI/Utils/Common/InputTextView.swift
```diff
@@ -169,6 +189,21 @@ class InputTextView: UITextView, AccessibilityView {
         placeholderLabel.isHidden = !text.isEmpty
     }
 
+    /// The height required to display the placeholder without truncation.
+    /// Returns `0` when the text view has content or the placeholder is hidden.
+    var placeholderFittingHeight: CGFloat {
```
martinmitrevski: feels a bit hacky to do all these calculations. Isn't there some more standard Apple ootb way of doing this?

nuno-vieira: Since there's no native placeholder API on `UITextView`, and we have a custom view for this, it will always going to be kinda hacky to do this. Either way,  I moved the manual height calculation into a standard `sizeThatFits(_:)` override on `InputTextView` so that the calculation is not that manual.

Sources/StreamChatSwiftUI/Utils/Common/InputTextView.swift
```diff
@@ -87,6 +94,15 @@ class InputTextView: UITextView, AccessibilityView {
...
     }
 
+    override open func didMoveToWindow() {
+        super.didMoveToWindow()
+        guard window != nil,
+              !didRequestInitialVoiceOverFocus,
+              UIAccessibility.isVoiceOverRunning else { return }
+        didRequestInitialVoiceOverFocus = true
+        UIAccessibility.post(notification: .screenChanged, argument: self)
+    }
```
laevandus, line 104: Just wondering about the state handling here. Text view should be added to the window only once, why there is a need to track it.

nuno-vieira: This is called everytime the channel view is presented or dismissed (onAppear), that is why the state

laevandus (approved):
Tried the placeholder wrapping. Placeholder text and input caret does not align perfectly when placeholder takes multiple lines, but not sure if worth it to adjust top spacing for alignment.

nuno-vieira:
> Tried the placeholder wrapping. Placeholder text and input caret does not align perfectly when placeholder takes multiple lines, but not sure if worth it to adjust top spacing for alignment.

For me works well 🤔 Either way, not very important at the moment, so lets skip this for now

PR #1503, opened 2026-06-25
Improve VoiceOver support for Polls

nuno-vieira:
### 🔗 Issue Links

- [IOS-1778](https://linear.app/stream/issue/IOS-1778/a11y-poll-header-not-identified-as-a-poll-just-text)
- [IOS-1779](https://linear.app/stream/issue/IOS-1779/a11y-poll-options-split-into-many-focus-stops-radiotextavatarcount)
- [IOS-1790](https://linear.app/stream/issue/IOS-1790/a11y-poll-results-decorative-trophyavatar-announced-as-separate-stops)

### 🎯 Goal

Navigating a poll with VoiceOver was confusing: the poll header read as plain text with no indication it was a poll, every option was fragmented into several focus stops (radio, text, avatar, count), and the results screen exposed decorative trophy/avatar images as separate stops. This PR makes polls announce as coherent, meaningful elements.

### 📝 Summary

- Announce the poll header as a single element identified as a poll, including the selection rule and option count.
- Combine each poll option into one focus stop that conveys its text, selection state, vote count, and position.
- Hide decorative trophy and voter-avatar images from VoiceOver and combine the results option headings and voter rows into single elements.
- Read the poll results question section as one element ("Question. <name>").

### 🛠 Implementation

- Poll header, options, results headings, and voter rows now use `accessibilityElement(children: .ignore)` with composed labels, and add the appropriate traits (`.isHeader`, `.isButton`).
- Poll options receive their position (`Option X of Y`) by passing the index and count down from the option lists.
- The label-building logic is centralised in a `PollAccessibility` helper so the branching (selection state, singular/plural/"including yours" vote wording, leading option, option position) can be unit tested directly.
- New localization keys were added for the poll accessibility announcements.

### 🧪 Manual Testing Notes

- Enable VoiceOver and open a channel with a poll message.
- Swipe through the poll: the header should announce as a poll with the rule and option count, and each option should be a single stop with its state, vote count, and position.
- Open Poll Results: the question, each option heading (including the leading option), and each voter row should each be a single, clean announcement with no separate "image" stops.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VoiceOver behavior for poll screens so headers, options, results, and voter rows are announced more clearly.
  * Added better accessibility support for poll options, including selection state, position, and vote count wording.
  * Fixed a layout issue that could freeze the app when viewing very long messages in the reactions overlay.

* **New Features**
  * Added new localized accessibility text for poll-related VoiceOver announcements.

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatMessageList/Polls/PollAttachmentView.swift
+++ b/Sources/StreamChatSwiftUI/ChatMessageList/Polls/PollAttachmentView.swift
@@ -60,14 +60,19 @@ public struct PollAttachmentView<Factory: ViewFactory>: View {
             .accessibilityAddTraits(.isHeader)
 
             VStack(spacing: tokens.spacingMd) {
-                ForEach(options.prefix(PollAttachmentViewModel.numberOfVisibleOptionsShown)) { option in
+                ForEach(
+                    Array(options.prefix(PollAttachmentViewModel.numberOfVisibleOptionsShown).enumerated()),
+                    id: \.element.id
+                ) { index, option in
                     PollOptionView(
                         viewModel: viewModel,
                         factory: factory,
                         option: option,
                         optionVotes: poll.voteCount(for: option),
                         maxVotes: poll.currentMaximumVoteCount,
-                        message: message
+                        message: message,
+                        optionIndex: index + 1,
+                        optionsCount: options.count
                     )
                     .layoutPriority(1) // do not compress long text
                 }
@@ -233,6 +238,10 @@ struct PollOptionView<Factory: ViewFactory>: View {
     /// If true, forces incoming color style for the radio button border and progress track,
     /// regardless of whether the message was sent by the current user.
     var forceIncomingStyle: Bool = false
+    /// The 1-based position of this option, used for the VoiceOver announcement.
+    var optionIndex: Int?
+    /// The total number of options, used for the VoiceOver announcement.
+    var optionsCount: Int?
 
     var body: some View {
         HStack(alignment: .top, spacing: tokens.spacingSm) {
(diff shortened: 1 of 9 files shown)


martinmitrevski (approved):
lgtm codewise. Although, the tests check was failing, I'm restarting it, so let's see.

Sources/StreamChatSwiftUI/ChatMessageList/Polls/PollAttachmentView.swift
```diff
@@ -283,6 +299,23 @@ struct PollOptionView<Factory: ViewFactory>: View {
         .onTapGesture {
             togglePollVote()
         }
+        .accessibilityElement(children: .ignore)
+        .accessibilityLabel(accessibilityLabel)
+        .accessibilityAddTraits(viewModel.poll.isClosed ? [] : .isButton)
+        .accessibilityAction {
+            guard !viewModel.poll.isClosed else { return }
```
laevandus: `onTapGesture` above does not have isClosed check. Align?

Sources/StreamChatSwiftUI/ChatMessageList/Polls/PollAttachmentView.swift
```diff
@@ -332,3 +365,90 @@ struct PollVotesIndicatorView: View {
...
+        let stateText = isSelected
+            ? L10n.Message.Polls.Accessibility.selected
+            : L10n.Message.Polls.Accessibility.notSelected
+        let votesText: String
+        if isSelected, voteCount > 1 {
+            votesText = L10n.Message.Polls.Accessibility.votesIncludingYours(voteCount)
+        } else if voteCount == 1 {
+            votesText = L10n.Message.Polls.voteSingular(voteCount)
+        } else {
+            votesText = L10n.Message.Polls.votes(voteCount)
```
laevandus: Existing problem, but we should have a ticket to use plural here instead of different keys because plural handling can be different in languages.

laevandus (approved):

Sources/StreamChatSwiftUI/ChatMessageList/Polls/PollAttachmentView.swift
nuno-vieira: Good catch — aligned both paths by moving the `isClosed` guard into `togglePollVote()`, so the row tap gesture and the accessibility action behave the same (and a closed poll can't be voted on through either). Done in 90abfd8f.

PR #1505, opened 2026-06-26
Improve VoiceOver announcements in the message list

nuno-vieira:
### 🔗 Issue Links

- [IOS-1777](https://linear.app/stream/issue/IOS-1777/a11y-messages-split-into-multiple-focus-stops-missing-sender-context)
- [IOS-1780](https://linear.app/stream/issue/IOS-1780/a11y-voice-message-attachments-not-announced-as-voice-messages)
- [IOS-1781](https://linear.app/stream/issue/IOS-1781/a11y-video-attachment-announced-as-image-leaks-raw-urlfilename)
- [IOS-1787](https://linear.app/stream/issue/IOS-1787/a11y-system-messages-not-identified-as-system-messages)
- [IOS-1788](https://linear.app/stream/issue/IOS-1788/a11y-deleted-message-is-split-into-multiple-focus-stops)
- [IOS-1795](https://linear.app/stream/issue/IOS-1795/a11y-translated-indicator-decorative-separator-announced)

### 🎯 Goal

VoiceOver navigation through the message list is noisy and lacks context: a single message splits into several focus stops (avatar, bubble, timestamp, delivery status), continuation messages drop the sender and time, and several content types are announced incorrectly (videos as images leaking raw URLs, voice messages as disconnected controls, deleted messages as two elements, system messages as plain text, and a decorative separator in the translated indicator). The goal is to announce each message as a single, meaningful element with full context.

### 📝 Summary

- Combine each message into a single VoiceOver element that announces sender, content, time, and (only on the last message in a sequence) delivery status.
- Always announce the sender and time, including on continuation messages, so jumping to any message keeps its context.
- Announce video attachments as videos with sender, optional duration, and time, without leaking the raw URL/filename.
- Announce voice messages as a single composite element with sender, duration, and time.
- Combine deleted messages into one "Message deleted" element and hide the decorative icon.
- Prefix system messages so they are identified as system messages.
- Hide the decorative separator in the translated indicator.

### 🛠 Implementation

- Add `MessageAccessibilityFormatter` to centralize the shared label building blocks (sender name, sent time, duration) and the attachment/voice labels, exposed via `Utils`.
- Add accessibility outputs to `MessageViewModel` (sender name, combined label, caption label, and a flag for whether bubble children should stay focusable) and combine elements in `MessageContainerView`.
- Update the media, voice, deleted, system, and annotation views to expose composite labels and hide decorative content.

### 🎨 Showcase

_VoiceOver-only change; no visual difference._

### 🧪 Manual Testing Notes

1. Enable VoiceOver and open a conversation.
2. Swipe through messages: each message should be a single focus stop announcing sender, content, time, and (on the last in a sequence) delivery status.
3. Navigate directly to a continuation message and confirm the sender and time are still announced.
4. Verify video attachments, voice messages, deleted messages, system messages, and translated messages each announce correctly as a single, clear element.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved VoiceOver accessibility for chat messages: clearer bubble announcements (sender, sent time, and delivery/read status when applicable), message actions, and consolidated labeling for captions and system messages.
  * Added richer attachment accessibility labels for images, videos, and voice recordings, including duration/timestamp variants and voice playback speed announcements.
  * Introduced new accessibility localization strings for these message and attachment announcements.

* **Bug Fixes**
  * Fixed VoiceOver duplication/mis-announcements by hiding decorative elements and making thumbnails/placeholders behave as intended for screen readers.

* **Tests**
  * Added coverage for message accessibility label formatting and attachment duration text behavior.


Sources/StreamChatSwiftUI/ChatMessageList/MessageContainerView.swift
```diff
@@ -101,52 +109,73 @@ struct MessageContainerView<Factory: ViewFactory>: View {
...
+                    }
+                )
+            )
+        }
+    }
+
+    @ViewBuilder
+    private var messageBubbleContent: some View {
+        Group {
+            if messageViewModel.keepsBubbleAccessibilityChildrenFocusable {
```
martinmitrevski: I would prefer if we avoid if-else here, and apply those accessibility changes with a ternary operator

Sources/StreamChatSwiftUI/ChatMessageList/MessageContainerView.swift
```diff
@@ -207,3 +237,19 @@ struct MessageContainerView<Factory: ViewFactory>: View {
         utils.messageListConfig
     }
 }
+
+/// The composite VoiceOver label (sender, content, time and delivery status)
+/// that a message's content should announce, so that focusing nested content
+/// (such as an attachment caption) reads the same thing as a message without
+/// attachments. `nil` when the surrounding message cell provides no composite
+/// label.
+struct MessageCompositeAccessibilityLabelKey: EnvironmentKey {
```
martinmitrevski, line 254: why do we need to pass it via the environment?

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
```diff
@@ -284,16 +285,66 @@ public struct MessageMediaAttachmentsContainerView<Factory: ViewFactory>: View {
...
+        // their view tree once the preview resolves (placeholder/spinner ->
+        // image + play indicator). If the accessibility element lived on this
+        // churning subtree, VoiceOver would rebuild and re-announce it after the
+        // preview loaded. Hide the visuals from accessibility and expose a single
+        // stable element that owns the label and activation instead.
+        .accessibilityHidden(true)
+        .overlay(mediaAccessibilityElement(for: item, index: index))
+    }
+
+    /// A stable accessibility element for a media cell, independent of the
```
martinmitrevski: hmm, feels a bit hacky?

martinmitrevski (commented):
ideally, we should avoid making structural changes to accommodate accessibility. Especially, not a fan of those if-else statements in the message view, that can impact performance.

nuno-vieira:
> ideally, we should avoid making structural changes to accommodate accessibility. Especially, not a fan of those if-else statements in the message view, that can impact performance.

Yes, I also wanted to avoid this, but due to our UI, there is no other option here. If we want to make Voice over work properly, we need to do some structural changes

Sources/StreamChatSwiftUI/ChatMessageList/AttachmentTextView.swift
```diff
@@ -7,6 +7,7 @@ import SwiftUI
 
 public struct AttachmentTextView<Factory: ViewFactory>: View {
     @Injected(\.tokens) private var tokens
+    @Environment(\.messageCompositeAccessibilityLabel) private var captionAccessibilityLabel
```
laevandus, line 10: If we would add an argument (e.g. `compositeAccessibilityLabel`) to the public init with default value (+ `AttachmentTextViewOptions`) then we can avoid environment key?

laevandus: I see, there are several layers. `MessageView` would need it as well.

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
```diff
@@ -284,16 +285,66 @@ public struct MessageMediaAttachmentsContainerView<Factory: ViewFactory>: View {
...
+            openMedia(at: index)
+        }
+        // The thumbnail loads asynchronously - video previews in particular swap
+        // their view tree once the preview resolves (placeholder/spinner ->
+        // image + play indicator). If the accessibility element lived on this
+        // churning subtree, VoiceOver would rebuild and re-announce it after the
+        // preview loaded. Hide the visuals from accessibility and expose a single
+        // stable element that owns the label and activation instead.
+        .accessibilityHidden(true)
+        .overlay(mediaAccessibilityElement(for: item, index: index))
```
laevandus, line 297: Would it be so bad without this?

nuno-vieira: Yes, it basically duplicates the announcement of the timestamp and the owner of the message :/

Sources/StreamChatSwiftUI/ChatMessageList/MessageContainerView.swift
nuno-vieira: Done

Sources/StreamChatSwiftUI/ChatMessageList/MessageContainerView.swift
nuno-vieira: Added a comment explaining why. TLDR: It would add a lot of noise otherwise, for something that is not absolute necessary

Sources/StreamChatSwiftUI/ChatMessageList/MessageMediaAttachmentsContainerView.swift
nuno-vieira: Added a comment explaning the approach. This is a common pattern, especially in UIKit

Sources/StreamChatSwiftUI/ChatMessageList/MessageContainerView.swift
```diff
@@ -50,8 +51,11 @@ struct MessageContainerView<Factory: ViewFactory>: View {
                         messageViewModel.topReactionsShown && messageViewModel.annotationsShown ? messageListConfig.messageDisplayOptions
                             .reactionsTopPadding(message) : 0
                     )
-                    .accessibilityElement(children: .contain)
                     .accessibilityIdentifier("MessageView")
+                    .environment(
+                        \.messageCompositeAccessibilityLabel,
+                        messageViewModel.captionAccessibilityLabel(showsAllInfo: showsAllInfo)
```
laevandus, line 56: Would it make sense to use the same terminology? messageViewModel.compositeAccessibilityLabel
I wonder if we wanna rethink about view model propagation in the next major, still pretty hard to send data further down the view hierarchy.

Sources/StreamChatSwiftUI/ChatMessageList/MessageViewModel.swift
```diff
@@ -98,6 +98,25 @@ import StreamChat
         message.isRightAligned
     }
 
+    /// Whether the message bubble contains elements that should remain separate
+    /// VoiceOver stops (attachments, polls, quotes, or failure affordances).
+    public var keepsBubbleAccessibilityChildrenFocusable: Bool {
```
laevandus, line 103: Might make sense to have it open, but we can do it later as well

laevandus (approved):
Some things are pretty complex. Env key does look cleaner than updating all of those inits. Maybe some things could be simpler when we could use accessibility representation for message view (iOS 15). Thinking about the stable identity thingy.

Sources/StreamChatSwiftUI/ChatMessageList/MessageViewModel.swift
nuno-vieira: I had it open initially, but for now better to be public only. Let's wait until someone actually needs to customize this.

PR #1508, opened 2026-06-30
Add video attachment disk cache

laevandus:
### 🔗 Issue Links

Resolves: [IOS-1800](https://linear.app/stream/issue/IOS-1800/)

Depends on: [GetStream/stream-chat-swift#4168](https://github.com/GetStream/stream-chat-swift/pull/4168)

### 🎯 Goal

Cache video attachments on disk while streaming, next play uses cached file

### 📝 Summary

- Added `VideoAttachmentCachingPolicy` for configuring video attachment disk caching from `MessageListConfig`.
- Stream cacheable video attachments through an `AVAssetResourceLoaderDelegate`, writing the streamed bytes to disk for later playback.
- Reuse completed cached files on later playback, validate cached videos before use, and fall back to the original remote URL when the custom cache loader cannot safely serve the response.
- Added unit coverage for video loader routing, cache policy configuration, and completed-file cache behavior.

### 🛠 Implementation

Video attachment caching is configured through `MessageListConfig.videoAttachmentCachingPolicy` and is disabled by default. When enabled, `StreamVideoPlayer.AVPlayerLoader` first checks `Utils.videoAttachmentDiskCache` for a completed local file. Cache misses use a custom-scheme `StreamVideoAsset` backed by `AVAssetResourceLoaderDelegate`, so playback and disk caching share a single network request.

The resource loader only handles responses it can safely serve from the sequential stream. It redirects unsupported responses, such as missing `Content-Length`, back to the original URL so playback can continue without caching. The loader also invalidates its `URLSession` on cancellation/completion and cancels asset loading when the video view disappears.

### Public API

- `VideoAttachmentCachingPolicy` — value type with `maxCacheSize` (total disk budget in bytes; `0` disables caching) and `allowedContentTypes` (defaults to `[.movie]`). Includes a `.disabled` preset.
- `MessageListConfig(videoAttachmentCachingPolicy:)` — controls video attachment disk caching. Disabled by default.

### 🧪 Manual Testing Notes

1. Open a channel with video attachment
2. Open the video attachment in the preview, let it play to the end
3. Close preview, turn off WI-FI, open preview again and play it
4. Video plays because it uses cached file

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/Generated/L10n.swift
laevandus, line 1117: Linter issue, I have slightly newer version locally.

Sources/StreamChatSwiftUI/Utils.swift
```diff
@@ -113,6 +115,10 @@ import StreamChatCommonUI
         self.mediaLoader = mediaLoader
         self.channelNameFormatter = channelNameFormatter
         self.avPlayerProvider = avPlayerProvider
+        diskCache = LRUDiskCache(
+            name: "com.getstream.StreamChatSwiftUI.diskCache",
+            maxSizeInBytes: diskCacheSizeInBytes
+        )
```
laevandus: Added it as general on disk cache which we can use for other things as well if there is a need.

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
```diff
@@ -342,31 +341,139 @@ struct StreamVideoPlayer: View {
...
+            completion: @escaping @MainActor (Result<AVPlayer, Error>) -> Void
+        ) {
+            avPlayerProvider.player(from: videoAsset) { result in
+                Task { @MainActor in
+                    completion(result)
+                }
+            }
+        }
+
+        private func prefetchToCache(key: String, fileExtension: String) {
```
laevandus: Way too complex to intercept streaming video to cache it, therefore caching happens separately with a cost of more bandwidth is used when using the media viewer (nothing is prefetched when in the message list).

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
```diff
@@ -33,6 +33,7 @@ import SwiftUI
         localLinkDetectionEnabled: Bool = true,
         isMessageEditedLabelEnabled: Bool = true,
         markdownSupportEnabled: Bool = false,
+        videoAttachmentCacheEnabled: Bool = true,
```
nuno-vieira: Do we want to enable it by default? I think it should be disabled by default since not many customers are interested in this 🤔

Sources/StreamChatSwiftUI/Utils/LRUDiskCache.swift
```diff
@@ -0,0 +1,140 @@
...
+            return nil
+        }
+
+        entries.withLock { entries in
+            guard var entry = entries?[name] else { return }
+            entry.lastUsedAt = Date()
+            entries?[name] = entry
+        }
+        return url
+    }
```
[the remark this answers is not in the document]
laevandus: It was an explicit decision for simplification purposes.

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
```diff
@@ -101,6 +103,10 @@ import SwiftUI
     public let localLinkDetectionEnabled: Bool
     public let isMessageEditedLabelEnabled: Bool
     public let markdownSupportEnabled: Bool
+    /// A boolean value indicating if video attachments should be cached on disk.
+    ///
+    /// It is enabled by default.
+    public let videoAttachmentCacheEnabled: Bool
```
nuno-vieira: Another approach could be to just have `videoAttachmentCachePolicy: VideoCachePolicy?`. Here, we would enable it if provided and set the max cache size.

I think that having a shared cache size for all types of resources could be a bit limiting IMO 🤔 I think it would be better to have the cache only for the videos and another for images etc... 

The cache policy approach is also nice because if we introduce other rules for the caching, like paralelel caching, or prefetch policies, we won't need to have breaking changes or introduce new types

Sources/StreamChatSwiftUI/Utils.swift
nuno-vieira: IMO I think it would better to have seperate cache sizes for different resources 🤔

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
```diff
@@ -342,31 +341,139 @@ struct StreamVideoPlayer: View {
...
+            self.cache = cache
+            self.cacheEnabled = cacheEnabled
+        }
+
+        func load(completion: @escaping @MainActor (Result<AVPlayer, Error>) -> Void) {
+            let canCache = cacheEnabled && !url.isFileURL && url.pathExtension.lowercased() != "m3u8"
+            let key = url.path
+            let fileExtension = url.pathExtension.isEmpty ? "mp4" : url.pathExtension
+
+            if canCache, let localURL = cache.cachedFileURL(forKey: key, fileExtension: fileExtension) {
```
nuno-vieira: `AVPlayerLoader` is `@MainActor`, so this `cache.cachedFileURL(...)` call runs synchronously on the main thread. On the first lookup that triggers `loadCacheIfNeeded()`, which scans the entire cache directory (up to 150 MB of entries) and reads each file size, plus a `FileManager.fileExists` check. Doing that synchronously on the main thread will block it and cause jank the first time a video is opened.

I would move the index load + lookup off the main thread — e.g. give `LRUDiskCache` its own I/O queue and an async `cachedFileURL` variant, then hop back to the main actor only with the result

nuno-vieira (commented):
Overall looks good 👍 Added a couple of suggestions

Sources/StreamChatSwiftUI/Utils.swift
laevandus: Made it video attachment specific.

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
laevandus: Done

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
laevandus: Added a queue

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
```diff
@@ -342,32 +342,148 @@ struct StreamVideoPlayer: View {
...
+        ) {
+            self.url = url
+            self.mediaLoader = mediaLoader
+            self.avPlayerProvider = avPlayerProvider
+            self.cache = cache
+            self.policy = policy
+            self.isPlayable = isPlayable
+        }
+
+        func load() async -> Result<AVPlayer, Error> {
```
martinmitrevski: why result, and not throwing function?

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
```diff
@@ -342,32 +342,148 @@ struct StreamVideoPlayer: View {
...
+                let asset = AVURLAsset(url: url)
+                asset.loadValuesAsynchronously(forKeys: ["playable"]) {
+                    continuation.resume(
+                        returning: asset.statusOfValue(forKey: "playable", error: nil) == .loaded && asset.isPlayable
+                    )
+                }
+            }
+        }
+
+        private func loadFromRemote() async -> Result<AVPlayer, Error> {
```
martinmitrevski: yes, there's no point here to wrap the return type in a result, the method below is already throwing

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
```diff
@@ -342,32 +342,148 @@ struct StreamVideoPlayer: View {
...
-                        switch result {
-                        case let .success(player):
-                            self.avPlayer = player
-                            try? AVAudioSession.sharedInstance().setCategory(.playback, options: [])
-                            self.avPlayer?.play()
-                        case let .failure(error):
-                            self.error = error
+                case let .success(fileRequest):
+                    let cache = self.cache
+                    URLSession.shared.downloadTask(with: fileRequest.urlRequest) { tempURL, response, error in
```
martinmitrevski: so we both play it and start a download? That doesn't seem very efficient. Can't we cache what was already streamed? Not realtime if it's complex, but maybe at the end?

Sources/StreamChatSwiftUI/Utils/LRUDiskCache.swift
```diff
@@ -0,0 +1,172 @@
...
+            .compactMap { url -> (name: String, entry: Entry)? in
+                let values = try? url.resourceValues(forKeys: [.fileSizeKey, .contentModificationDateKey])
+                guard let size = values?.fileSize else { return nil }
+                return (url.lastPathComponent, Entry(size: size, lastUsedAt: values?.contentModificationDate ?? Date()))
+            }
+        entries = Dictionary(uniqueKeysWithValues: scanned ?? [])
+    }
+
+    private static func storageName(forKey key: String, fileExtension: String?) -> String {
+        let hex = SHA256.hash(data: Data(key.utf8)).map { String(format: "%02x", $0) }.joined()
```
martinmitrevski: what's the idea behind this?

Sources/StreamChatSwiftUI/Utils/LRUDiskCache.swift
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    private func loadCacheIfNeeded() {
+        guard entries == nil else { return }
+        let scanned = (try? FileManager.default.contentsOfDirectory(
+            at: directory,
+            includingPropertiesForKeys: [.fileSizeKey, .contentModificationDateKey],
+            options: [.skipsHiddenFiles]
+        ))?
+            .compactMap { url -> (name: String, entry: Entry)? in
```
martinmitrevski: weird formatting here

Sources/StreamChatSwiftUI/Utils/LRUDiskCache.swift
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            self.entries?[name] = Entry(size: size, lastUsedAt: Date())
+            var cacheSize = self.entries?.values.reduce(0, { $0 + $1.size }) ?? 0
+            var entryCount = self.entries?.count ?? 0
+            let namesToEvict = (self.entries ?? [:])
+                .sorted(by: { $0.value.lastUsedAt < $1.value.lastUsedAt })
+                .prefix { entry in
+                    guard cacheSize > self.maxSizeInBytes, entryCount > 1 else { return false }
+                    cacheSize -= entry.value.size
```
martinmitrevski: this can go negative?

Sources/StreamChatSwiftUI/Utils/LRUDiskCache.swift
```diff
@@ -0,0 +1,172 @@
...
+                completion(nil)
+                return
+            }
+
+            self.loadCacheIfNeeded()
+            let name = Self.storageName(forKey: key, fileExtension: fileExtension)
+            let destination = self.directory.appendingPathComponent(name)
+
+            do {
+                try FileManager.default.createDirectory(at: self.directory, withIntermediateDirectories: true)
```
martinmitrevski: this whole method feels too big and hard to reason about - in some places we use helper methods, but in others, like here, we do all the file manipulation stuff that can easily go into a separate method.

Sources/StreamChatSwiftUI/Utils/LRUDiskCache.swift
```diff
@@ -0,0 +1,172 @@
...
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import CryptoKit
+import Foundation
+import StreamChat
+
+/// A thread-safe, size-bounded disk cache with least-recently-used (LRU) eviction.
+/// All file I/O and in-memory index access are serialized on a private queue.
+final class LRUDiskCache: @unchecked Sendable {
```
martinmitrevski: shouldn't this logic go into the common module, so UIKit benefits from it?

martinmitrevski (commented):
we should address some stuff here before merging

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
laevandus: AVFoundation does not give access to the data it streamed. Only way seems to be handling the streaming and caching while playing. I am gonna try this out and see how complex does it get.

laevandus: I tried it out, but it gets pretty long. 
https://github.com/GetStream/stream-chat-swiftui/compare/video-attachment-cache-streaming?expand=1

Sources/StreamChatSwiftUI/Utils/AVPlayerLoader/StreamVideoCache.swift
```diff
@@ -0,0 +1,159 @@
...
+        directory.appendingPathComponent(Self.storageName(forKey: key, fileExtension: fileExtension))
+    }
+
+    private static func storageName(forKey key: String, fileExtension: String?) -> String {
+        let hex = SHA256.hash(data: Data(key.utf8)).map { String(format: "%02x", $0) }.joined()
+        if let fileExtension, !fileExtension.isEmpty {
+            return "\(hex).\(fileExtension)"
+        }
+        return hex
+    }
```
laevandus: For file system safe file name

Sources/StreamChatSwiftUI/Utils/LRUDiskCache.swift
laevandus: Key is taken from url path which can be anything with custom CDN. SHA makes sure it is suitable as file name.

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
```diff
@@ -342,31 +342,124 @@ struct StreamVideoPlayer: View {
...
+                avPlayer?.play()
+            } catch {
+                self.error = error
+            }
+        }
+    }
+}
+
+extension StreamVideoPlayer {
+    @MainActor final class AVPlayerLoader {
```
martinmitrevski: probably better if we move this into its own file?

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
```diff
@@ -142,6 +149,30 @@ import SwiftUI
     public var supportedMessageActions: @MainActor (SupportedMessageActionsOptions) -> [MessageAction]
 }
 
+/// A policy describing how video attachments are cached on disk.
+public struct VideoAttachmentCachingPolicy: Sendable, Equatable {
```
martinmitrevski: maybe final class would be better here?

martinmitrevski (commented):
looks good, also did some tests, but best if Alexey does one round. Also, maybe we should increase the test coverage a bit here?

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
laevandus: Changed it to final class

Sources/StreamChatSwiftUI/ChatMessageList/Gallery/MediaViewer.swift
laevandus: Moved to StreamAVPlayerLoader.swift

laevandus:
> looks good, also did some tests, but best if Alexey does one round. Also, maybe we should increase the test coverage a bit here?

Added tests for the player loader which was missing before.

martinmitrevski (approved):
LGTM ✅  Maybe you can create a follow up ticket to eventually bring this to the common module?

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
```diff
@@ -142,6 +149,34 @@ import SwiftUI
...
+    public init(maxCacheSize: Int64, allowedContentTypes: Set<UTType> = [.movie]) {
+        self.maxCacheSize = maxCacheSize
+        self.allowedContentTypes = allowedContentTypes
+    }
+}
+
+public extension VideoAttachmentCachingPolicy {
+    /// Video attachment caching is disabled. This is the default.
+    static let disabled = VideoAttachmentCachingPolicy(maxCacheSize: 0)
+}
```
nuno-vieira: Why not just make it a nullable config? 🤔

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
```diff
@@ -142,6 +149,34 @@ import SwiftUI
     public var supportedMessageActions: @MainActor (SupportedMessageActionsOptions) -> [MessageAction]
 }
 
+/// A policy describing how video attachments are cached on disk.
+public final class VideoAttachmentCachingPolicy: Sendable {
+    /// The maximum total size of the video attachment disk cache, in bytes.
+    ///
+    /// Caching is disabled when this is `<= 0`.
```
nuno-vieira: Kinda odd, not very "swifty" 🤔

nuno-vieira (approved):
LGTM, just added 2 nits

nuno-vieira:
> LGTM ✅ Maybe you can create a follow up ticket to eventually bring this to the common module?

Ah yes, this makes sense, UIKit could get this for free in the future. If there are no breaking changed in moving this later, fine, otherwise, we should do this now

Sources/StreamChatSwiftUI/ChatMessageList/MessageListConfig.swift
laevandus: Done

laevandus:
I will refactor this to extract the streaming and loader part to common UI.

StreamChatSwiftUI.xcodeproj/project.pbxproj
```diff
@@ -1534,8 +1534,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/GetStream/stream-chat-swift.git";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 5.6.0;
+				kind = branch;
+				branch = "feature/cachingavplayer";
```
laevandus: Don't merge

laevandus:
Verified again with common UI integration that cache gets picked up.

PR #1510, opened 2026-06-30
Fix link preview images not loading

laevandus:
### 🔗 Issue Links

[IOS-1806](https://linear.app/stream/issue/IOS-1806)

### 🎯 Goal

Link attachment previews stopped showing their image — the title rendered but the preview image never loaded, even when `previewURL` was a valid, reachable URL.

### 📝 Summary

- Fix `StreamAsyncImage` never starting the image load when the content closure renders nothing for non-`.success` phases (which is what link previews do).
- Add a regression test.

### 🛠 Implementation

`StreamAsyncImage` attached its load `.task` directly to the caller's `content` view. `LinkAttachmentView`'s closure only renders a view for `.success`, so on the initial `.empty` phase the content resolved to an empty view — and SwiftUI never fires `.task`/`.onAppear` on a view that resolves to nothing. The load therefore never started and the phase stayed `.empty` forever (a deadlock: the content only materializes on `.success`, but `.success` requires the load). Wrapping the content in a `ZStack` gives the `.task` a stable container that always materializes. This is a fix at the shared view, so it covers every caller whose content renders nothing outside `.success`, not just link previews.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
| Simulator Screenshot - iPhone 17 Pro - 2026-06-30 at 16 38 54 | Simulator Screenshot - iPhone 17 Pro Max - 2026-06-30 at 16 38 47 |

### 🧪 Manual Testing Notes

Open/send a message with a link whose preview has an image (e.g. `https://zebra-coffee.com/`). The preview image now renders in the bubble.

### ☑️ Contributor Checklist


nuno-vieira (approved):
LGTM! ✅ Nice catch!

PR #1515, opened 2026-07-02
Send Files/iCloud picker images and videos as their proper type

testableapple:
### 🔗 Issue Links

Resolve https://linear.app/stream/issue/IOS-1818

### 🎯 Goal

Images and videos picked from the Files/iCloud document picker were always sent as `type: file`, so they rendered as downloadable files everywhere instead of inline. The Photos/Media picker was unaffected because it builds assets with a proper type. This makes the document-picker path behave consistently.

### 📝 Summary

- Resolve the attachment type from the picked file's extension instead of hardcoding `.file`, so images and videos are sent as `image`/`video` and render inline.
- Audio files continue to be treated as regular files (the composer only supports audio through voice recordings).

### 🛠 Implementation

- In `MessageAttachmentsConverter.assetsToPayloads(_:)`, derive the type via StreamChat's existing `AttachmentType(fileExtension:)` (resolves image/video/audio/file from the MIME type), mirroring the UIKit SDK's `ComposerVC.documentPicker(_:didPickDocumentsAt:)`.
- `AnyAttachmentPayload(localFileURL:attachmentType:)` already builds the correct payload from a local URL — the same path the Photos picker uses — so the upload pipeline supports it unchanged.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
| Image from Files sent as a file attachment | Image from Files rendered inline |

### 🧪 Manual Testing Notes

1. Open a channel, tap the attachment button, choose **Files** / iCloud.
2. Pick an image (and a video). Send.
3. The image/video renders inline in the message list, on web, and on other clients (previously appeared as a downloadable file).

### ☑️ Contributor Checklist

- [x] This change should be manually QAed
- [x] Changelog is updated with client-facing changes
- [x] New code is covered by unit tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Images and videos chosen via the Files/iCloud picker are now correctly recognized and sent as inline media (not as downloadable attachments).
  * Video metadata and display sizing are preserved so playback sizing is more accurate.
* **Tests**
  * Expanded composer attachment tests to cover async media processing and verify correct image/video/document payloads using generated temporary files.
* **Documentation**
  * Updated the upcoming changelog entry with this fix.

PR #1516, opened 2026-07-02
Send Files/iCloud picker images and videos as their proper type (v4)

testableapple:
### 🔗 Issue Links

Resolve https://linear.app/stream/issue/IOS-1818

### 🎯 Goal

Images and videos picked from the Files/iCloud document picker were always sent as `type: file`, so they rendered as downloadable files everywhere instead of inline. The Photos/Media picker was unaffected because it builds assets with a proper type. This makes the document-picker path behave consistently.

### 📝 Summary

- Resolve the attachment type from the picked file's extension instead of hardcoding `.file`, so images and videos are sent as `image`/`video` and render inline.
- Audio files continue to be treated as regular files (the composer only supports audio through voice recordings).

### 🛠 Implementation

- In `MessageAttachmentsConverter.assetsToPayloads(_:)`, derive the type via StreamChat's existing `AttachmentType(fileExtension:)` (resolves image/video/audio/file from the MIME type), mirroring the UIKit SDK's `ComposerVC.documentPicker(_:didPickDocumentsAt:)`.
- `AnyAttachmentPayload(localFileURL:attachmentType:)` already builds the correct payload from a local URL — the same path the Photos picker uses — so the upload pipeline supports it unchanged.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
| Image from Files sent as a file attachment | Image from Files rendered inline |

### 🧪 Manual Testing Notes

1. Open a channel, tap the attachment button, choose **Files** / iCloud.
2. Pick an image (and a video). Send.
3. The image/video renders inline in the message list, on web, and on other clients (previously appeared as a downloadable file).

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
@@ -642,12 +642,7 @@ open class MessageComposerViewModel: ObservableObject {
     /// `MessageAttachmentsConverter` still resolves the correct attachment type from the file
     /// extension when sending it.
     private static func mediaAsset(fromPickedFileURL url: URL) -> AddedAsset? {
-        let didStartAccessing = url.startAccessingSecurityScopedResource()
-        defer {
-            if didStartAccessing {
-                url.stopAccessingSecurityScopedResource()
-            }
-        }
+        _ = url.startAccessingSecurityScopedResource()
 
         switch AttachmentType(fileExtension: url.pathExtension) {
         case .image:
@@ -991,14 +986,9 @@ open class MessageComposerViewModel: ObservableObject {
     
     private func checkAttachmentSize(with url: URL?) -> Bool {
         guard let url = url else { return true }
-
-        let didStartAccessing = url.startAccessingSecurityScopedResource()
-        defer {
-            if didStartAccessing {
-                url.stopAccessingSecurityScopedResource()
-            }
-        }
-
+        
+        _ = url.startAccessingSecurityScopedResource()
+        
         do {
             let fileSize = try AttachmentFile(url: url).size
             let canAdd = fileSize < chatClient.maxAttachmentSize(for: url)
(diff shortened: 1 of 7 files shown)


testableapple:
Note: the equivalent v5 PR (#1515) offloads the picked-file copy + thumbnail extraction off the main actor to avoid blocking the UI on large files. On this v4 branch the pick-time routing is kept **synchronous** intentionally — the view model here is not `@MainActor` and the routing lives in `addedFileURLs.didSet`, so an async offload would be a riskier, non-symmetric change on a maintenance branch. The blocking only affects very large picked files; keeping v4 minimal.

testableapple:
Update on approach: this v4 branch now resolves the attachment type **at send** (`AttachmentType(fileExtension:)` in `assetsToPayloads`) rather than routing picked media into composer assets at pick time. Reason: the pick-time routing had to hook `addedFileURLs.didSet`, which reroutes *every* mutation of that property (drafts/edits included, not just picker additions) and broke several unrelated attachment tests. On a maintenance branch the send-time fix is the minimal, low-risk way to fix the customer's actual issue (images/videos sent as `file`). The inline-composer-preview improvement (routing at pick time, off-main media building) stays a v5-only change (#1515).

martinmitrevski (approved):
why's the fix different between versions?

testableapple:
> why's the fix different between versions?

@martinmitrevski, both versions send the correct type now, but v5 also works in composer preview, while v4 only shows the content on sent. I could not make it work as expected on v4, so i went with a safer and smaller fix.

Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
```diff
@@ -602,6 +603,84 @@ open class MessageComposerViewModel: ObservableObject {
...
+    }
+
+    /// Builds a media asset for an image or video picked from the Files/iCloud picker,
+    /// mirroring how camera/Photos picks are already added as inline media. Returns `nil`
+    /// for other files, or if the media can't be read (e.g. a corrupted or not-yet-downloaded
+    /// iCloud file), in which case the caller falls back to a generic file attachment;
+    /// `MessageAttachmentsConverter` still resolves the correct attachment type from the file
+    /// extension when sending it.
+    private static func mediaAsset(fromPickedFileURL url: URL) -> AddedAsset? {
+        _ = url.startAccessingSecurityScopedResource()
```
laevandus: Defer { url.stop... }? Leaks resources without stop.

testableapple: I fixed it everywhere, please review:
- v4: https://github.com/GetStream/stream-chat-swiftui/pull/1516/commits/0c3955ec6771ba1234cfa1a3772ccb57e6eb49d6
- v5: https://github.com/GetStream/stream-chat-swiftui/pull/1515/commits/e2d2d36590822d609a86e090931dc9570de04987

laevandus (approved):

PR #1527, opened 2026-07-08
Render picked images/videos inline in the composer on v4

nuno-vieira:
### 🔗 Issue Links

Related to https://linear.app/stream/issue/IOS-1818

### 🎯 Goal

#1516 fixed images/videos picked from the Files/iCloud picker being **sent** with the wrong attachment type, but they still render as a generic file chip in the composer itself, unlike camera/Photos picks. This PR adds the same inline preview that already exists on `develop`, adapted to v4's composer architecture.

### 📝 Summary

- `MessageComposerViewModel.addFileURLs(_:)` now builds an inline media asset (with a synchronous thumbnail) for picked images/videos, the same way the camera and Photos pickers already add them, so they render inline instead of as a file chip.
- Other files (and media that can't be read, e.g. an undownloaded iCloud file) still fall back to a regular file attachment; `MessageAttachmentsConverter` resolves the correct type from the extension when sending, so it's still sent as the right type either way.
- Only newly picked files are routed through this — draft/edit restoration and attachment removal continue to mutate `addedFileURLs` directly and are unaffected.

### 🛠 Implementation

v4's composer doesn't have a single entry point for picker additions — `addedFileURLs: [URL]` is a plain `@Published` array bound directly to the native file picker, and also reused for draft/edit restoration and removal. A previous attempt at this fix hooked into `addedFileURLs.didSet`, which fired for **all** of those cases (not just new picks) and was reverted because it broke draft/edit restoration and unrelated tests.

This PR avoids that by not touching `addedFileURLs`'s own setter at all. Instead:
- `MessageComposerViewModel.pickedFileURLsBinding` is a new binding that reads `addedFileURLs` and, on write, diffs the newly appended URLs and routes them through `addFileURLs(_:)`.
- `MessageComposerView` and the demo's `AppleMessageComposerView` now pass this binding (instead of `$viewModel.addedFileURLs`) to `makeAttachmentPickerView(...)`.

Since draft restoration (`updateComposerAssets`) and removal (`removeAttachment(with:)`) mutate `addedFileURLs` directly in Swift code rather than through this SwiftUI binding, they're completely unaffected — the only thing that writes through `pickedFileURLsBinding` is the native file picker appending newly picked URLs.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
| Image from Files rendered as a file chip in the composer | Image from Files rendered inline in the composer |

### 🧪 Manual Testing Notes

1. Open a channel, tap the attachment button, choose **Files** / iCloud.
2. Pick an image and a video.
3. Both should render inline in the composer (thumbnail, not a file chip), and inline in the message list after sending.
4. Sanity check that removing an attachment, restoring a draft with a file attachment, and editing a message with an existing file attachment all still work as before.

### ☑️ Contributor Checklist

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
@@ -664,16 +664,18 @@ open class MessageComposerViewModel: ObservableObject {
                 at: CMTimeMake(value: 0, timescale: 1),
                 actualTime: nil
             ) else { return nil }
-            let duration = CMTimeGetSeconds(asset.duration)
+            let durationSeconds = CMTimeGetSeconds(asset.duration)
+            let duration: TimeInterval? = durationSeconds.isFinite && !durationSeconds.isNaN ? durationSeconds : nil
             let naturalSize = asset.tracks(withMediaType: .video).first?.naturalSize ?? .zero
             return AddedAsset(
                 image: UIImage(cgImage: cgImage),
                 id: UUID().uuidString,
                 url: url,
                 type: .video,
+                extraData: duration.map { ["duration": .string($0.composerVideoDurationString)] } ?? [:],
                 originalWidth: Double(naturalSize.width),
                 originalHeight: Double(naturalSize.height),
-                duration: duration.isFinite ? duration : nil
+                duration: duration
             )
         default:
             return nil
--- a/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
@@ -2,6 +2,7 @@
 // Copyright © 2026 Stream.io Inc. All rights reserved.
 //
 
+import AVFoundation
 import Combine
 import Photos
 import StreamChat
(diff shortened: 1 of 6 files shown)


Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
```diff
@@ -602,6 +603,82 @@ open class MessageComposerViewModel: ObservableObject {
...
+    }
+
+    /// Builds a media asset for an image or video picked from the Files/iCloud picker,
+    /// mirroring how camera/Photos picks are already added as inline media. Returns `nil`
+    /// for other files, or if the media can't be read (e.g. a corrupted or not-yet-downloaded
+    /// iCloud file), in which case the caller falls back to a generic file attachment;
+    /// `MessageAttachmentsConverter` still resolves the correct attachment type from the file
+    /// extension when sending it.
+    private static func mediaAsset(fromPickedFileURL url: URL) -> AddedAsset? {
+        _ = url.startAccessingSecurityScopedResource()
```
laevandus, line 645: Defer with stop method? Docs say it leaks kernel resources if stop is not called.

Sources/StreamChatSwiftUI/ChatChannel/Composer/MessageComposerViewModel.swift
```diff
@@ -602,6 +603,82 @@ open class MessageComposerViewModel: ObservableObject {
...
+                originalWidth: Double(image.size.width * scale),
+                originalHeight: Double(image.size.height * scale)
+            )
+        case .video:
+            let asset = AVURLAsset(url: url, options: nil)
+            let imageGenerator = AVAssetImageGenerator(asset: asset)
+            imageGenerator.appliesPreferredTrackTransform = true
+            guard let cgImage = try? imageGenerator.copyCGImage(
+                at: CMTimeMake(value: 0, timescale: 1),
+                actualTime: nil
```
laevandus, line 665: Fast enough for not having to be async?

PR #1529, opened 2026-07-09
Improve channel list rendering performance

nuno-vieira:
### 🔗 Issue Links

- [IOS-912](https://linear.app/stream/issue/IOS-912/channel-list-synchronize-takes-a-lot-of-time-when-multiple-members-per)
- Depends on GetStream/stream-chat-swift#4165

### 🎯 Goal

Reduce main-thread hangs while scrolling and paginating the channel list.

### 📝 Summary

- ~~Diff channel updates instead of replacing the whole array, so `ForEach` only reconciles rows that actually changed.~~
- Resolve a channel row's index in O(1) via a lookup map instead of scanning the array on every row appearance.
- Render the channel list divider as a row overlay instead of a separate sibling node.
- Use the channel list query's default page size when paginating, and only trigger pagination within the last 8 rows.

### 🛠 Implementation

Channel updates are applied with `Array.difference(from:)` so unchanged rows keep their identity, and row index resolution is O(1). Pagination fires less eagerly (within the last 8 rows, default page size). Row visuals are unchanged.

> Note: We initially explored backing the channel list with a native SwiftUI `List` on iOS 15+ for true row virtualization, but reverted it since it conflicts with custom swipe-gesture implementations in customized `ViewFactory`s. The channel list keeps rendering with `LazyVStack`/`ScrollView`.

> Note: The `StreamChat` dependency temporarily points at the `perf/channel-list-performance-improvements` branch (GetStream/stream-chat-swift#4165). It should be pointed back at a released version once that PR is merged.

### 🎨 Showcase

With all these fixes, including LLC changes, we can drop from around 70 hangs when scrolling fast through new pages to just 14. In order to completely drop the hangs to 0, the only way is actually to bypass the database. I've done a prototype, and only without the CoreData processing can I get a complete 0 hangs.

#### Before

image-7602a2aa-c70a-48e6-a8bc-e77447d02c94

#### After
image-ab44ddcc-ef22-4426-aa5d-f37f68790008

### 🧪 Manual Testing Notes

Scroll a large channel list and paginate; scrolling should stay smooth without main-thread hangs. Row visuals, dividers, selection and swipe actions should be unchanged.

### ☑️ Contributor Checklist


Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
```diff
@@ -567,7 +567,13 @@ import UIKit
...
 
     private func updateChannels() {
-        channels = controller?.channels ?? [ChatChannel]()
+        let newChannels = controller?.channels ?? [ChatChannel]()
+        // Apply the diff between the current and new channels so unchanged
+        // elements keep their identity and position. This lets `ForEach`
+        // reconcile only the rows that actually moved, were inserted, or were
+        // removed instead of rebuilding every row on each update.
+        let difference = newChannels.difference(from: channels)
+        channels = channels.applying(difference) ?? newChannels
```
laevandus: `ChatChannel` implements `Identifiable` which is the `cid.rawValue`, so does difference from then actually improve diffing on the SwiftUI side? If I understood correctly then this helps only if identity is the pointer of the object, but if we have our own id implementation, then it does not help much. Could you help me out if I have incorrect impression.

nuno-vieira: In theory I was thinking that this should have made no difference in terms of performance as well. Even if it has, it is minimal, so I reverted this for now 👍

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
```diff
@@ -83,27 +84,16 @@ public struct ChannelList<Factory: ViewFactory>: View {
...
-    var channels: [ChatChannel]
-    @Binding var selectedChannel: ChannelSelectionInfo?
-    @Binding var swipedChannelId: String?
-    private var onItemTap: @MainActor (ChatChannel) -> Void
-    private var onItemAppear: @MainActor (Int) -> Void
-    private var channelDestination: (@MainActor (ChannelSelectionInfo) -> Factory.ChannelDestination)?
-    private var trailingSwipeRightButtonTapped: @MainActor (ChatChannel) -> Void
-    private var trailingSwipeLeftButtonTapped: @MainActor (ChatChannel) -> Void
-    private var leadingSwipeButtonTapped: @MainActor (ChatChannel) -> Void
+    private var content: ChannelList<Factory>
```
martinmitrevski: why do we keep the view as a var, this doesn't seem like a standard thing to do in SwiftUI? Any benefits?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
```diff
@@ -83,27 +84,16 @@ public struct ChannelList<Factory: ViewFactory>: View {
...
-            leadingSwipeButtonTapped: leadingSwipeButtonTapped
+            leadingSwipeButtonTapped: leadingSwipeButtonTapped,
+            currentUserId: chatClient.currentUserId,
+            showChannelListDividerOnLastItem: utils.channelListConfig.showChannelListDividerOnLastItem
         )
     }
 }
 
 /// LazyVStack displaying list of channels.
 public struct ChannelsLazyVStack<Factory: ViewFactory>: View {
```
martinmitrevski, line 92: is this view still even used?

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelList.swift
```diff
@@ -117,65 +107,145 @@ public struct ChannelsLazyVStack<Factory: ViewFactory>: View {
...
+    let onItemAppear: @MainActor (Int) -> Void
+    let channelDestination: (@MainActor (ChannelSelectionInfo) -> Factory.ChannelDestination)?
+    let trailingSwipeRightButtonTapped: @MainActor (ChatChannel) -> Void
+    let trailingSwipeLeftButtonTapped: @MainActor (ChatChannel) -> Void
+    let leadingSwipeButtonTapped: @MainActor (ChatChannel) -> Void
+    let currentUserId: UserId?
+    let showChannelListDividerOnLastItem: Bool
+
+    var body: some View {
+        let channelIndexLookup = channelListIndexLookup(for: channels)
```
martinmitrevski: hmm, I don't understand what's the point here? This lookup gets recreated on every body evaluation. If it only because of the onAppear check? I don't think that improves performance, at least from my measurements in the past - even for large lists.

Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
```diff
@@ -182,13 +182,13 @@ import UIKit
     public func checkForChannels(index: Int) {
         handleChannelAppearance()
 
-        if index < (controller?.channels.count ?? 0) - 15 {
+        if index < (controller?.channels.count ?? 0) - 8 {
```
martinmitrevski: with this change, the first time it doesn't paginate straight away, I need to bounce few times before it starts - we should still tweak this

martinmitrevski (commented):
I'm not sure if these changes have a big impact on performance tbh. I think the LLC ones contribute more, can you share some more data?

martinmitrevski (approved):
Nice performance boost!

Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(url: "https://github.com/GetStream/stream-chat-swift.git", from: "5.6.0")
+        .package(url: "https://github.com/GetStream/stream-chat-swift.git", branch: "perf/channel-list-performance-improvements")
```
laevandus: Reminder to change before merge.

laevandus (approved):

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
```diff
@@ -389,11 +399,24 @@ import SwiftUI
...
             return
         }
-        let attachments = try? convertAddedAssetsToPayloads()
         let mentionedUserIds = mentionedUsers.map(\.id)
         let availableCommands = channelController.channel?.config.commands ?? []
         let command = availableCommands.first { composerCommand?.id == "/\($0.name)" }
 
+        // Skips the resave (new local id + network request) when the composer
+        // still matches the existing draft, e.g. the channel was opened and
+        // closed again without any edits.
```
nuno-vieira:
```suggestion
```

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
```diff
@@ -201,6 +203,12 @@ import SwiftUI
     )
     
     private var timer: Timer?
+
+    /// `true` while ``fillDraftMessage()`` is populating the composer from an
+    /// existing draft, so the resulting `text`/asset assignments don't trigger
+    /// typing events as if the user had typed them.
```
nuno-vieira:
```suggestion
```

PR #1530, opened 2026-07-14
Add support for rendering messages from the top of the list

nuno-vieira:
### 🔗 Issue Links

Resolves [IOS-1058](https://linear.app/stream/issue/IOS-1058/swiftuibumbledailymotionbitnance-support-message-ordering-from-the-top)

### 🎯 Goal

Add support for rendering messages from the top of the list.

### 📝 Summary

- Add `MessageListConfig.shouldMessagesStartAtTheTop` (defaults to `false`, fully opt-in)
- When enabled and the messages are shorter than the visible list, messages are laid out from the top with the remaining space acting as an implicit bottom spacer
- Once messages grow past the visible height, the list behaves exactly like the existing scrollable inverted list, including regular scroll/insertion animations
- Add a "Messages Start at the Top" toggle to the demo app's App Configuration screen

### 🛠 Implementation

- The message stack is given a `minHeight` equal to the available list height and bottom-aligned; after the list's upside-down flip this places the messages at the top
- The available height is measured synchronously via a `GeometryReader` wrapping the list content, so the top-aligned frame never lags behind the real viewport size on first layout
- Once the messages outgrow the viewport the `minHeight` clamp has no effect by construction, so the list naturally degrades to the regular inverted behavior with no extra state to track
- While top-aligned, animations are suppressed for the whole scroll view on content updates, since animating the row insertion and the spacer's resize as two separate animations made existing messages look like they were shifting instead of a new message simply appearing

### 🧷 Scroll stability while rubber-banding

With few messages, dragging the top-aligned list up rubber-bands the rows under the navigation bar. When the rows are hosted in a `LazyVStack`, that overscroll triggers a layout invalidation mid-gesture: the scroll view clamps the drag and the messages visibly jump back. The lazy container was isolated empirically as the cause — the jump reproduces with every scroll callback and preference handler in the list disabled, only with two or more messages, and never with an eager container. The exact trigger is internal to the lazy layout (it sizes and places children relative to the visible region, and an overscrolled, flipped, height-clamped viewport puts that machinery in an edge case), so the safe conclusion is behavioral: eager rendering of the visible rows keeps the layout stable during rubber-banding; a lazy one doesn't. Two solutions were explored; both fix the jump and both were validated with the same drag/hold E2E scenarios (including live message events arriving mid-drag).

**1. Container swap — currently used**

The message stack renders as an eager `VStack` while top-aligned with fewer than 20 messages, and as the regular `LazyVStack` otherwise. 20 messages always overflow the viewport, so beyond that top alignment (and the jump) is irrelevant.

- ✅ Simple and easy to reason about: short lists are fully eager, long lists are exactly the existing lazy list
- ✅ No laziness bookkeeping at all while the feature is visually active
- ⚠️ Crossing the 20-message threshold swaps the container type, which resets the view identity of the whole stack once (a one-time reflow with no row state to lose in practice)
- ⚠️ Eager rendering cost is bounded but real if rows are heavy (attachments), though capped at 20 rows

**2. Eager chunk inside the lazy stack — explored and reverted (`d3737ed7`, reverted in `ba0d5e85`)**

The container stays a `LazyVStack` always; the newest 20 rows are grouped into a single plain `VStack` child, with older history remaining individual lazy rows. A lazy stack manages its children per direct child, so the chunk is laid out as one opaque unit with no lazy-layout boundaries inside it — which was verified to keep the layout stable during overscroll, same as the full container swap.

- ✅ No container swap: the outer stack's identity is stable at every message count
- ✅ Older history in long channels stays lazy; crossing the threshold only re-identifies one row at a time (the row leaving the chunk) instead of the whole list
- ⚠️ Hackier structure: it relies on the (undocumented but long-standing) per-child realization semantics of lazy stacks, and splits the rows across two `ForEach`es with a nested container in between
- ⚠️ Every new message migrates the 20th row from the eager chunk to the lazy `ForEach`, so its structural identity churns continuously as the conversation grows

The container swap was kept because its complexity is front-loaded and obvious (one `if` around the container), while the chunk approach spreads subtle identity implications across the list body. If the one-time reflow at the threshold ever becomes noticeable, the chunk approach is a validated drop-in alternative.

### 🎨 Showcase

Few messages render at the top of the list instead of the bottom, with the remaining space kept empty below them.

### 🧪 Manual Testing Notes

- Enable "Messages Start at the Top" in the demo app's App Configuration screen
- Open a channel with few messages and confirm they render at the top, not the bottom
- Send messages until the list fills the screen and confirm it smoothly transitions to normal scrollable behavior
- Open/close the keyboard and the attachment picker and confirm the messages never shift or scroll under the navigation bar
- Scroll/rubber-band the top-aligned list up and down and confirm the drag never clamps or jumps, including while another participant sends a message mid-drag
- Repeat the above with the glass/floating composer style enabled

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an option to start message lists at the top when conversations contain few messages.
  * Added a Message List setting in the demo app to enable or disable top-aligned messages.
  * Preserved appropriate spacing and scrolling behavior for short and long conversations.

* **Documentation**
  * Documented the new message list configuration option in the upcoming changelog.


Sources/StreamChatSwiftUI/ChatMessageList/MessageListView.swift
```diff
@@ -187,8 +200,16 @@ public struct MessageListView<Factory: ViewFactory>: View, KeyboardReadable {
                         Color.clear.preference(key: ScrollViewOffsetPreferenceKey.self, value: offset)
                         Color.clear.preference(key: WidthPreferenceKey.self, value: width)
                     }
+                    // A GeometryReader has a 10-point ideal height inside a
+                    // vertical ScrollView. The top-aligned stack already fills
+                    // the container, so that height would create phantom
+                    // overflow and allow the messages to scroll under the
+                    // navigation bar.
+                    .frame(height: topAlignedMinHeight != nil ? 0 : nil)
```
laevandus: Should it also touch alignment?

I haven't done full investigation, but my 5 min research showed that alignment is also useful

```
GeometryReader { proxy in
    ScrollView {
        VStack {
            // content
        }
        .frame(
            maxWidth: .infinity,
            minHeight: proxy.size.height,
            alignment: .bottom / .top
        )
    }
}
```

nuno-vieira: We are doing that in TopAlignedFillModifier

martinmitrevski: feels a bit hacky imo

Sources/StreamChatSwiftUI/ChatMessageList/MessageListView.swift
```diff
@@ -232,17 +253,18 @@ public struct MessageListView<Factory: ViewFactory>: View, KeyboardReadable {
...
-                                        showsLastInGroupInfo: showsLastInGroupInfo,
-                                        showUnreadSeparator: showUnreadSeparator,
-                                        showThreadRepliesSeparator: showThreadRepliesSeparator
-                                    ) :
-                                    additionalTopPadding(
-                                        showsLastInGroupInfo: showsLastInGroupInfo,
-                                        showUnreadSeparator: showUnreadSeparator,
-                                        showThreadRepliesSeparator: showThreadRepliesSeparator
-                                    )
+                                (message == messages.last && topAlignedMinHeight != nil ? tokens.spacingXxs : 0)
```
martinmitrevski: wow, this is quite complex, let's split it up - there are ternary operators, additions, can't easily tell what's the idea

Sources/StreamChatSwiftUI/ChatMessageList/MessageListView.swift
```diff
@@ -361,19 +391,33 @@ public struct MessageListView<Factory: ViewFactory>: View, KeyboardReadable {
                         if !shouldJump {
                             return
                         }
+                        if topAlignedMinHeight != nil {
+                            var transaction = Transaction()
+                            transaction.disablesAnimations = true
```
martinmitrevski: this is fine for this case

Sources/StreamChatSwiftUI/ChatMessageList/MessageListView.swift
```diff
@@ -561,6 +645,47 @@ struct ScrollTargetLayoutModifier: ViewModifier {
...
+/// `MessageListContentHeightPreferenceKey`. Only tracked when `enabled`, so
+/// consumers who don't use `shouldMessagesStartAtTheTop` pay no extra cost.
+struct MessageListContentHeightTrackingModifier: ViewModifier {
+    var enabled: Bool
+
+    @ViewBuilder
+    func body(content: Content) -> some View {
+        if enabled {
+            content.background(
+                GeometryReader { proxy in
```
martinmitrevski: hmm, another geometry reader - wonder if it can impact performance?

martinmitrevski (commented):
fine with the code changes if it's the only way - but @testableapple should run a solid QA. Also, would be nice to measure the performance, we put another geometry reader, plus some additional ifs, anchors etc.

Sources/StreamChatSwiftUI/ChatMessageList/MessageListView.swift
nuno-vieira: Yes any solution we provide will be a bit hacky

Sources/StreamChatSwiftUI/ChatMessageList/MessageListView.swift
nuno-vieira: It won't impact because this is only used in the beginning. After the message list has enough messages, it switches back to normal rendering

Sources/StreamChatSwiftUI/ChatMessageList/MessageListView.swift
nuno-vieira: This was already like this before, but I agree this is already too much, so refactored it ✅

laevandus (approved):
I did some testing and works well. The initial shift of the whole message list is fixed as well. Only other alternative I could think of was dropping flipped handling for bottom alignment, but then many VStacks and message index handling logic needs to be updated as well (messagingOrder would change and then index logic does not match anymore). That seems like a much bigger change than what we have here. Therefore, I feel like we should go with this implementation if we don't find any other layout issues during testing.

martinmitrevski (approved):
LGTM ✅ important one for @testableapple, both cases should be tested - starting from top and bottom

martinmitrevski (approved):
lgtm

PR #1546, opened 2026-07-27
Fix unresponsive media picker when there is a lot of iCloud assets

nuno-vieira:
### 🔗 Issue Links

[IOS-1801](https://linear.app/stream/issue/IOS-1801/attachment-gallery-picker-gets-stuck-when-there-are-multiple-videos)

### 🎯 Goal

When the photo library had a lot of assets in iCloud, the attachment media picker was effectively unusable:
- No attachment could be selected until every visible item finished downloading and processing.
- Scrolling stuttered, and thumbnails kept flickering.
- After scrolling far down, closing and reopening the picker froze the app for seconds.

The goal is a picker that is interactive immediately, scrolls smoothly at any speed, and opens/closes instantly regardless of library size — deferring the expensive work to the moment the user actually picks an asset, like WhatsApp does.

### 📝 Summary

- Assets are no longer downloaded and processed when an item appears; the full asset is fetched when it is tapped.
- Local assets are resolved on tap, with images converted to JPG — fast enough that no spinner is needed, like before.
- iCloud assets are downloaded on tap, with a spinner while the download runs.
- Oversized videos are compressed on tap, including ones downloaded from iCloud, with a spinner until the item is ready.
- Thumbnails are served by a caching image manager with a bounded cache, and requests are cancelled when items scroll away or the picker is dismissed.
- The grid no longer touches the Photos database while SwiftUI updates it, which removes the multi-second hang when closing/opening the picker.
- Improves overall test coverage of the attachment picker

#### Issues found and root causes

**1. Selection blocked until every visible video was processed** (the reported bug)

`AttachmentMediaPickerItemView.onAppear` called `asset.requestContentEditingInput` with `isNetworkAccessAllowed = true` for every visible item, and compressed the result when it exceeded the maximum attachment size. Every visible video therefore downloaded from iCloud and ran an `AVAssetExportSession` in parallel, just from scrolling past it.

**2. Stuttering scroll and flickering thumbnails**

Thumbnails lived in a single `@Published var loadedImages` dictionary on `PhotoAssetLoader`, which every item observed through `@ObservedObject`. Each completed load published a change that invalidated *every* visible item, so N loads caused N full re-renders of the visible grid. Requests went to `PHImageManager.default()` with `.highQualityFormat` and were never cancelled, so a fast scroll queued a backlog of full-quality decodes for items that were already gone.

**3. Unbounded memory growth**

`loadedImages` never evicted anything, so scrolling through a large library retained every decoded thumbnail for the lifetime of the loader.

**4. Multi-second main-thread hang when closing/opening the picker** (the worst one)

Three things compounded:

- The picker stays in the view hierarchy while hidden (it is animated out via height/offset/opacity), so nothing is ever torn down.
- `LazyVGrid` keeps every row it has instantiated, so after scrolling to the bottom of a large library thousands of rows were live.
- The grid was keyed by asset (`ForEach(collection)`, `PHAsset: Identifiable`), so SwiftUI's `ForEach` bookkeeping subscripted `PHFetchResultCollection` for every instantiated row on every update. `PHFetchResult` only keeps a small batch window materialized — every access outside it rebuilds `PHAsset` objects from the Photos store.

The result was thousands of Photos database fetches on the main thread per update. Profiling the hang shows exactly that, with `ForEachState.update(view:)` at the top of the Photos work:

```
396.00 ms  13,7%   objc_msgSend
 91.00 ms   3,1%    _PFRawIndexForKnownKey
 91.00 ms   3,1%     -[_PFResultObject objectForKey:]
 19.00 ms   0,7%      -[PHAsset initWithFetchDictionary:propertyHint:photoLibrary:]
                       ...
 19.00 ms   0,7%                 -[PHFetchResult objectAtIndex:]
 19.00 ms   0,7%                  PHFetchResultCollection.subscript.getter
 17.00 ms   0,6%                     ForEachState.update(view:)
```

**5. Every composer update re-evaluated every instantiated item**

Items take closures (`onImageTap`, `imageSelected`), which SwiftUI cannot compare, so it could not skip any of them. This only hurts when re-evaluating an item is expensive; with the on-appear/on-tap split below, item bodies do no Photos work, so re-evaluation is cheap and no diffing workaround is needed.

**6. Reopening the picker invalidated the whole grid**

`askForPhotosPermission()` runs on every presentation and `fetchAssets()` always assigned a brand-new `PHFetchResult`, so every item was invalidated even when the library had not changed.

**7. Thumbnail requests outlived the picker**

Requests were only cancelled from an item's `onDisappear`, which does not run when the picker is merely hidden, and a fast scroll coalesces the grid's appearance callbacks. Scrolling quickly and dismissing left the remaining requests running.

#### The new flow

Appearing an item only loads its thumbnail; everything else waits for a tap. A tap first tries a local-only resolve, which is fast and needs no spinner, and only falls back to an iCloud download — with the spinner — when the asset is not on the device.

```mermaid
flowchart TD
    A([Item appears]) --> B[Load thumbnail<br/>cached, cancellable, no network]

    G([Item tapped]) --> H{Already selected,<br/>or resolved?}
    H -- yes --> I([Select immediately])
    H -- no --> J[Resolve file URL<br/>local only, no spinner]
    J --> C{Available<br/>on device?}
    C -- no --> D[Download from iCloud<br/>with spinner]
    C -- yes --> K{Video above<br/>max attachment size?}
    D --> K
    K -- yes --> L[Compress<br/>with spinner]
    K -- no --> M([Select])
    L --> M
```

Images are delivered as JPG (originals are usually HEIC); the conversion is fast, so local photos select without any spinner, like before. The spinner only appears while downloading from iCloud or compressing a video, and an item is only selected once it is fully prepared. If a compressed video is still above the limit, the composer's existing size check rejects it and shows the "Max size" alert.

#### Changes per file

- **`AttachmentMediaPickerItemView`** (with `MediaPickerAssetHandler`): owns its thumbnail state and releases it when it scrolls away, so off-screen items do not retain images outside the loader's cache. Resolves and compresses on tap — local first without a spinner, then an iCloud download with one — and cancels its pending request on disappear. No longer an `@ObservedObject` observer of the loader.
- **`AttachmentMediaPickerView`**: index-keyed `ForEach` so grid updates never re-materialize the fetch result, with items keyed by asset identifier so their state resets when a different asset lands on an index; cancels all in-flight thumbnail requests when hidden and scrolls to the top when shown.
- **`PhotoAssetLoader`**: `PHCachingImageManager` with `.opportunistic` delivery, `.fast` resizing and no network access for thumbnails; a bounded `NSCache` (200 objects / 64 MB, cost based on decoded size) that is cleared on memory warnings; per-asset tracking of in-flight requests with `cancelImageLoad(for:)` and `cancelAllImageLoads()`.
- **`MessageComposerViewModel`**: keeps the current `PHFetchResult` when a refetch has the same count and the same first/last asset identifiers.
- **`AttachmentPickerView`, `DefaultViewFactory`, `AttachmentMediaPickerViewOptions`**: pass the picker's `isDisplayed` state down to the media picker. `AttachmentMediaPickerViewOptions.isDisplayed` is a new property and the new initializer parameter is defaulted, so this is source compatible.

### 🎨 Showcase

| Before | After |
|--------|--------|
|  <video src="https://github.com/user-attachments/assets/998903c6-3d6a-4190-bd00-b4da12987b59"/> | <video src="https://github.com/user-attachments/assets/574bd497-0075-4a62-a40c-d0ba76ca9a22"/> |

### 🧪 Manual Testing Notes

Use a device with a large photo library (thousands of assets) containing several large, iCloud-optimised videos.

1. Open the attachment picker: the grid is interactive immediately, and no video downloads while it is idle.
2. Scroll to the bottom as fast as possible: scrolling stays smooth and thumbnails fill in without flickering.
3. From the bottom, close and reopen the picker several times: no freeze, and the grid starts back at the top.
4. Tap a photo: it is selected instantly.
5. Tap an iCloud video: a spinner runs while it downloads (and compresses, if it is above the maximum attachment size), then it becomes selected. Tapping it again deselects instantly.
6. Tap a video that is still too large after compression: the "Max size" alert is shown and the video is not attached.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the media attachment picker becoming unresponsive and slow to open when photo libraries contain many iCloud assets.
  * Improved picker responsiveness with cache-first thumbnail loading and more dependable cancellation when the picker is dismissed/hidden.
  * Reduced unnecessary media grid refreshes by detecting unchanged photo-library content before re-rendering.

* **Tests**
  * Added/expanded picker and asset loader test coverage for thumbnail caching, cancellation, asset URL resolution, and image/video selection with compression and size limits.


CHANGELOG.md
```diff
@@ -4,8 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 # Upcoming
 
 ### 🐞 Fixed
+- Fix media attachment picker being unresponsive and slow to open with large photo libraries [#1546](https://github.com/GetStream/stream-chat-swiftui/pull/1546)
```
nuno-vieira:
```suggestion
- Fix media attachment picker being unresponsive and slow with many iCloud assets [#1546](https://github.com/GetStream/stream-chat-swiftui/pull/1546)
```

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/AttachmentMediaPickerItemView.swift
```diff
@@ -12,44 +12,38 @@ public struct AttachmentMediaPickerItemView: View {
...
     var selectedAssetIds: Set<String>?
-    
-    private var assetType: AssetType {
-        asset.mediaType == .video ? .video : .image
-    }
 
     public init(
         assetLoader: PhotoAssetLoader,
-        requestId: PHContentEditingInputRequestID? = nil,
+        requestId _: PHContentEditingInputRequestID? = nil,
```
laevandus: We should collect these cleanups somewhere for v6. This argument should be removed then. Let's create a general ticket for public interface argument cleanup?

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/AttachmentMediaPickerItemView.swift
```diff
@@ -12,44 +12,38 @@ public struct AttachmentMediaPickerItemView: View {
...
         self.selectedAssetIds = selectedAssetIds
     }
- 
+
     public var body: some View {
         let selected = isAssetSelected(asset.localIdentifier)
+        let image = handler.currentImage
         ZStack {
-            if let image = assetLoader.loadedImages[asset.localIdentifier] {
+            if let image {
```
laevandus: `if let image = handler.currentImage`?

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/AttachmentMediaPickerView.swift
```diff
@@ -87,6 +109,39 @@ public struct AttachmentMediaPickerView: View {
     }
 }
 
+/// Defers `PHAsset` resolution until the body runs, and skips that body for off-screen
+/// rows via `Equatable` so composer updates do not hit the Photos database.
+private struct MediaPickerCellView: View, Equatable {
```
laevandus: Wondering if wrapping `PHAsset` with another type and handling this there is better than making the view equatable. 
E.g. `StreamPHAsset` which is property of the cell view. Could possibly avoid index based handling. Did you consider that approach?

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
```diff
@@ -6,50 +6,120 @@ import Combine
...
 
-    /// Loads an image from the provided asset.
-    func loadImage(from asset: PHAsset) {
-        if loadedImages[asset.localIdentifier] != nil {
+    // Bounded so that scrolling through a large library does not retain every
+    // decoded thumbnail. NSCache also evicts automatically under memory pressure.
+    private let imageCache: NSCache<NSString, UIImage> = {
+        let cache = NSCache<NSString, UIImage>()
+        cache.countLimit = 200
+        cache.totalCostLimit = 64 * 1024 * 1024
```
laevandus: Image loader currently only has 50 MB limit. Wondering if 64 MB is too much and should it get flushed more aggressively. 64 was picked because it gave good enough performance?

laevandus (commented):
I did some testing and seems to work well. Picker feels a little bit snappier compared to develop (I can't reproduce the initial performance issue with my phone so the difference looks smaller).

nuno-vieira:
> I did some testing and seems to work well. Picker feels a little bit snappier compared to develop (I can't reproduce the initial performance issue with my phone so the difference looks smaller).

In order to reproduce the performance issue, you need to pick the scroll indicator and drag it to the very bottom of the gallery, then close and reopen the picker multiple times. This may also be worse in my condition where I have a lot of images and videos on iCloud.

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/AttachmentMediaPickerItemView.swift
nuno-vieira: For now, I've added 2 inits, one deprecated, and the new one

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/AttachmentMediaPickerItemView.swift
nuno-vieira: Fixed

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
nuno-vieira: Changed to 50MB to be consistent with image loader 👍

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/AttachmentMediaPickerView.swift
nuno-vieira: I've tried it, but we need the index based approach to avoid accessing the PHAsset and causing the performance issue

martinmitrevski: I think the current solution is fine, no need to provide a wrapper

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
```diff
@@ -6,50 +6,136 @@ import Combine
...
             return
         }
 
+        let assetId = asset.localIdentifier
+        cancelImageLoad(for: asset)
+
         let options = PHImageRequestOptions()
         options.version = .current
-        options.deliveryMode = .highQualityFormat
+        options.deliveryMode = .opportunistic
```
martinmitrevski, line 72: why are all these properties changed?

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
```diff
@@ -6,50 +6,136 @@ import Combine
...
+        options.resizeMode = .fast
+        options.isNetworkAccessAllowed = false
```
martinmitrevski, line 74: if this is false, does it mean that we need to explicitly load it?

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
```diff
@@ -65,6 +151,44 @@ import SwiftUI
...
+        completion: @escaping @MainActor (URL?) -> Void
+    ) -> PHImageRequestID {
+        let options = PHImageRequestOptions()
+        options.version = .current
+        options.isNetworkAccessAllowed = allowsNetworkAccess
+
+        // Marked Sendable so the handler makes no isolation assumption about the delivery queue.
+        return imageManager.requestImageDataAndOrientation(for: asset, options: options) { @Sendable data, dataUTI, _, _ in
+            // Photos delivers the data on the main thread, so writing the file is moved off it.
+            Task.detached(priority: .utility) {
```
martinmitrevski: shall we use a higher priority than utility?

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
```diff
@@ -90,7 +214,23 @@ import SwiftUI
 
     /// Clears the cache when there's memory warning.
     func didReceiveMemoryWarning() {
-        loadedImages = [String: UIImage]()
+        imageCache.removeAllObjects()
+    }
+
+    /// Writes image data to a temporary JPG file, converting it first when it is not JPG already.
+    nonisolated static func temporaryJpgURL(for data: Data, dataUTI: String?) -> URL? {
```
martinmitrevski, line 223: did we do this already or is it a new thing? If new, why do we need it?

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
```diff
@@ -894,10 +894,23 @@ import SwiftUI
...
+            if let current = imageAssets, Self.haveSameContent(current, assets) {
+                return
+            }
+            imageAssets = assets
         }
     }
 
+    static func haveSameContent(_ lhs: PHFetchResult<PHAsset>, _ rhs: PHFetchResult<PHAsset>) -> Bool {
+        guard lhs.count == rhs.count else { return false }
+        return lhs.firstObject?.localIdentifier == rhs.firstObject?.localIdentifier
```
martinmitrevski, line 910: hm, is this condition enough?

Sources/StreamChatSwiftUI/ViewFactory/Options/AttachmentViewFactoryOptions.swift
```diff
@@ -474,17 +474,21 @@ public final class AttachmentMediaPickerViewOptions: Sendable {
     public let isAssetSelected: @MainActor (String) -> Bool
     /// Snapshot of the currently selected asset identifiers.
     public let selectedAssetIds: [String]?
-    
+    /// Whether the picker is currently being displayed.
+    public let isDisplayed: Bool
```
martinmitrevski, line 478: I don't really understand, why do you need this property? Isn't it easier to do the cleanup onDisappear of the picker itself?

martinmitrevski (commented):
looks good, also tested a bit and scrolling is smooth (I didn't have any icloud backed items though). Would be nice if @testableapple also tests, since we have a lot of code changes. On the topic of code changes - seems a lot, maybe you can check if everything is still used? (in case you tried different approaches)

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
nuno-vieira: If this is not changed, when scrolling too fast (when dragging the scroll indicator) a dummy placeholder is rendered instead of the actual image, and it looks a bit bad. This way, the system adapts the resolution depending on the scrolling, and it automatically selects the best resolution depending on the conditions.

You can see on the video in the PR description that the placeholders appear, instead of the actual images.

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
nuno-vieira: If the image is in iCloud, then when we select the image, you see a brief loading spinner and then the image is selected. This is to avoid downloading multiple images and videos at the same time if they are on iCloud. So only when the user selects this asset is it downloaded from iCloud.

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/AttachmentMediaPickerView.swift
nuno-vieira: Seems like this has no real impact actually, so I removed it 👍

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
nuno-vieira: Good point. This task runs on two paths: the on-appear prefetch (nobody waiting) and the on-tap resolve, where the user is watching a spinner while a HEIC photo gets decoded and re-encoded to JPG. Bumped it to `.userInitiated` in 3880ad1f — the prefetch path only handles the few visible cells, so the contention risk is negligible.

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
nuno-vieira: It's the existing HEIC-to-JPG conversion, moved and improved rather than a new behavior. Before, `requestContentEditingInput` gave us a file URL and on tap we read the file, decoded it to `UIImage` and re-encoded it with `saveAsJpgToTemporaryUrl()` on the main thread — even when it was already JPG. The replacement API (`requestImageDataAndOrientation`) returns raw `Data` instead of a file URL, and the composer needs a local file URL, so this helper writes the data to a temp file: directly when it is already JPG (skipping the decode/re-encode), converting only when it is HEIC/PNG, and off the main thread.

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
nuno-vieira: It should be good enough. When an asset is added or removed, the count is changed, and the first or last item are also different. To be 100% accurate, it would be better to use `PHPhotoLibraryChangeObserver`, but I would like to avoid that for now, and make this simpler

Sources/StreamChatSwiftUI/ViewFactory/Options/AttachmentViewFactoryOptions.swift
nuno-vieira: `onDisappear` never fires here: the picker is dismissed by collapsing its height to 0 (`height: viewModel.overlayShown ? popupSize : 0` in `MessageComposerView`), so it, and every grid cell, stays in the view hierarchy. That's also why `AttachmentPickerView` already used `isDisplayed` + `onChange` for the permissions prompt instead of `onAppear` — this PR just forwards the existing flag one level down. It additionally drives scrolling the grid back to the top when the picker is shown again, which `onAppear`/`onDisappear` can't do for a view that is never removed.

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/MediaPickerAssetHandler.swift
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    // MARK: - Private
+
+    private func loadThumbnail() {
+        guard thumbnail == nil, assetLoader.cachedImage(for: asset) == nil else { return }
+        assetLoader.loadImage(for: asset, targetSize: CGSize(width: 250, height: 250)) { [weak self] image in
+            self?.thumbnail = image
+        }
+    }
```
laevandus, line 89: I guess no point of complicating it and passing the actual frame size to thumbnail.

Sources/StreamChatSwiftUI/ChatComposer/AttachmentPicker/AttachmentMediaPicker/PhotoAssetLoader.swift
laevandus: It is a good change to delay downloading assets. 👍

laevandus (approved):
I did some additional testing and visually it gives much better result without seeing these placeholder images flashing while scrolling or mostly just seeing them when dragging from the scrollbar.
Left one more minor comment about the 250x250 thumbnail size. Would have been nice to not introduce custom view equality, but not many ways to do this differently when we have closures as properties there.

Sources/StreamChatSwiftUI/ChatComposer/MessageComposerViewModel.swift
martinmitrevski: hmm, but if you modify an asset, e.g. crop it, that change won't be reflected? I'm fine with that, just so we know the limitations.

martinmitrevski (approved):
LGTM ✅

PR #1549, opened 2026-07-29
Remove divider from the message actions

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1917/v5-ui-polish-message-action-menu-shows-separators-between-every-item.

### 🎯 Goal

Remove divider from the message actions per the design.

### 📝 Summary

### 🛠 Implementation

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Style / UI**
  * Refined message action menu separators so non-destructive actions are grouped without dividers, while a divider is still shown before destructive actions.
* **Tests**
  * Updated snapshot coverage for message action layouts to validate: single destructive action, grouped destructive actions, and non-destructive-only lists.
* **Documentation**
  * Updated the upcoming changelog entry to reflect the message action separator behavior change.


laevandus (approved):
✅

Sources/StreamChatSwiftUI/ChatMessageList/Reactions/MessageActions/MessageActionsView.swift
```diff
@@ -57,8 +57,6 @@ public struct MessageActionsView: View {
                         }
                         .accessibilityLabel(action.title)
                     }
-
-                    Divider()
```
nuno-vieira, line 61: According to Figma, it looks like destructive actions should have the divider, and should be at the end 🤔 

Image

nuno-vieira (changes requested):

Sources/StreamChatSwiftUI/ChatMessageList/Reactions/MessageActions/MessageActionsView.swift
martinmitrevski: updated now

Sources/StreamChatSwiftUI/ChatMessageList/Reactions/MessageActions/MessageActionsView.swift
```diff
@@ -84,4 +88,26 @@ public struct MessageActionsView: View {
...
+        if let lastRegularIndex = actions.lastIndex(where: { !$0.isDestructive }) {
+            let destructiveGroupIndex = lastRegularIndex + 1
+            dividerIndex = destructiveGroupIndex < actions.count ? destructiveGroupIndex : nil
+        }
+
+        return actions.enumerated().map { index, action in
+            ActionRow(action: action, showsDivider: index == dividerIndex)
+        }
+    }
+}
```
nuno-vieira, line 106: This works fine, but not very customizable. Would be better if this either stayed on the `MessageActionsViewModel` or the `ChannelInfoViewModel`. Since the latter is the one responsible for creating the actions, it would probably make sense to extend the `MessageAction` with `showsDivider`, and the info view model would make the logic to show or not show the divider. 

An even better approach would be to have a `MessageActionItemViewModel` so that the UI logic would be decoupled from the actual MessageAction. But I believe this would require some breaking changes, most likely. 

Either way, none are blocking at the moment.

nuno-vieira (approved):
Looks good, could be improved in terms of customiziblity, but not a blocker

PR #1550, opened 2026-07-29
Fix thread list items being lighter than the background in dark mode

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1911/v5-ui-polish-thread-list-items-use-wrong-elevation-token-in-dark-mode.

### 🎯 Goal

In dark mode, thread list items rendered lighter than the app background, so the list looked like a raised surface floating on the screen instead of part of it.

### 📝 Summary

- Thread list items and the thread list background no longer appear lighter than the app background in dark mode.

### 🛠 Implementation

`makeThreadListBackground` and `makeThreadListItemBackground` used `colors.backgroundCoreElevation1`, which is defined as `UIColor(light: chrome0, dark: chrome50)` — a raised surface in dark mode. Both now use `colors.backgroundCoreElevation0` (`chrome0` in both appearances), which matches the app background and is the same token the channel list already uses for its items.

Light mode is pixel-identical, since `backgroundCoreElevation0` and `backgroundCoreElevation1` both resolve to `chrome0` there.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
|  img   |  img  |

### 🧪 Manual Testing Notes

Run the demo app in dark mode and open the thread list — the rows should blend into the screen background rather than appearing as a lighter block. Verify light mode is unchanged.

### ☑️ Contributor Checklist

## Bug Fixes

- Fixed dark-mode thread list styling so thread list items no longer appear lighter than the app background.
- Improved contrast and consistency across thread list backgrounds.

## Tests

- Updated thread list UI snapshot coverage for both light and dark appearances at the standard screen size.


Sources/StreamChatSwiftUI/ViewFactory/DefaultViewFactory.swift
```diff
@@ -985,15 +985,15 @@ extension ViewFactory {
...
+        Color(options.colors.backgroundCoreElevation0)
             .edgesIgnoringSafeArea(.bottom)
     }
 
     public func makeThreadListItemBackground(
         options: ThreadListItemBackgroundOptions
     ) -> some View {
         let colors = InjectedValues[\.colors]
-        return Color(colors.backgroundCoreElevation1)
+        return Color(colors.backgroundCoreElevation0)
```
nuno-vieira, line 996: DOn't we need to re-record the snapshots?

martinmitrevski: seems no, everything passes. Do we have tests for the thread list in dark mode?

nuno-vieira: Yeah seems like we are not using our Internal AssertSnapshot, so dark mode is not covered. It would be nice to add it

martinmitrevski: will add them

martinmitrevski: @nuno-vieira updated, there are some new snapshots now

nuno-vieira (approved):

martinmitrevski:
@testableapple the recording snapshots action doesn't seem to be working

laevandus (approved):

PR #1551, opened 2026-07-30
Fix tab bar covering the composer during the back swipe

martinmitrevski:
### 🔗 Issue Links

Resolves https://linear.app/stream/issue/IOS-1915/v5-ui-polish-floating-nav-bar-leaks-into-message-view-during-swipe.

### 🎯 Goal

In apps that host the channel list inside a `TabView`, slowly swiping back from the channel view made the tab bar appear on top of the still-visible channel view, covering the composer. The tab bar should only come back once the swipe has finished and the channel list is on screen.

### 📝 Summary

- The tab bar no longer appears over the pushed channel view while the interactive back swipe is in progress; it shows up after the swipe completes.
- A cancelled back swipe keeps the tab bar hidden, instead of leaving it on top of the composer.
- Tapping the back button no longer shows the tab bar during the pop animation.
- `TabBarAccessor` now owns the tab bar visibility handling, so the channel list view no longer keeps UIKit state.

### 🛠 Implementation

The tab bar was hidden imperatively (`tabBar.isHidden`) while a channel is open, and shown again from `ChatChannelListViewModel.checkTabBarAppearance()`, which the channel list calls on `onItemAppear`. During a back swipe the channel list items re-appear while the gesture is still running, so the tab bar was un-hidden mid-transition. Since the tab bar is drawn above the navigation controller's content, it landed on top of the channel view and its composer.

`TabBarAccessor` is now responsible for applying the visibility, and it tracks whether its own screen is on screen using the standard view controller appearance callbacks:

- Hiding is applied immediately.
- Showing is applied only while the screen is visible, otherwise it is deferred to `viewDidAppear`, which runs when the pop has actually finished.
- A cancelled interactive pop delivers `viewWillDisappear`/`viewDidDisappear` instead of `viewDidAppear`, so the tab bar simply stays hidden until the next completed pop.

Because of that, `ChatChannelListView` no longer needs `@State` for the tab bar, the `onReceive` subscription, or the `UIKit` import — it just passes the desired state (`TabBarAccessor(isTabBarHidden: viewModel.hideTabBar)`). The `UITabBarAppearance` reconfiguration that ran on every visibility change turned out to be unnecessary and was removed. The behaviour is the same on all supported iOS versions, and `MessageListConfig.handleTabBarVisibility == false` still opts out entirely.

### 🎨 Showcase

Screenshots taken mid-gesture (finger held at ~65% of the drag), using a temporary `TabView` harness in the UI test app:

| Before | After |
| ------ | ----- |
| Tab bar visible at the bottom, overlapping the channel view's composer (the `+` button sits behind it) | No tab bar; the composer stays fully visible, and the tab bar appears once the finger is lifted and the pop finishes |

### 🧪 Manual Testing Notes

In the SwiftUI demo app (channel list inside a `TabView`):

1. Open a channel.
2. Slowly swipe from the left edge and hold the finger mid-screen — the tab bar should not appear, and the composer should stay fully visible.
3. Release to complete the swipe — the tab bar appears together with the channel list.
4. Repeat, but release early so the swipe is cancelled — the tab bar should stay hidden while you are back in the channel, including after new messages arrive.
5. Go back with the back button — the tab bar should appear after the pop animation, not during it.

### ☑️ Contributor Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the tab bar could cover the message composer when swiping back to the channel list.
  * Improved tab bar visibility management as screens appear, disappear, and reappear.
* **UI Changes**
  * Removed separators between message action menu items, keeping a single separator before destructive actions.
* **Tests**
  * Added coverage for tab bar visibility across navigation and screen lifecycle scenarios.


StreamChatSwiftUI.xcodeproj/project.pbxproj
```diff
@@ -263,6 +263,7 @@
 				Tests/ChatChannelList/MoreChannelActionsView_Tests.swift,
 				Tests/ChatChannelList/MoreChannelActionsViewModel_Tests.swift,
 				Tests/ChatChannelList/SearchResultsView_Tests.swift,
+				Tests/ChatChannelList/TabBarAccessor_Tests.swift,
```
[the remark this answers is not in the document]
martinmitrevski: are you sure about that? I see it in the test target.

StreamChatSwiftUITests/Tests/ChatChannelList/TabBarAccessor_Tests.swift
```diff
@@ -0,0 +1,129 @@
+//
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+@testable import StreamChatSwiftUI
+import XCTest
+
+@MainActor final class TabBarAccessor_Tests: XCTestCase {
```
nuno-vieira, line 8: Does it make sense to keep these tests? Feels like this is mostly something only manually tested

nuno-vieira (approved):

StreamChatSwiftUITests/Tests/ChatChannelList/TabBarAccessor_Tests.swift
martinmitrevski: it doesn't harm, especially if we make changes to the hidden handling. But yes, should always be manually tested as well.

PR #1552, opened 2026-07-30
Fix lighter edge ring in the message actions scrim blur

martinmitrevski:
### 🔗 Issue Links

[IOS-1914](https://linear.app/stream/issue/IOS-1914/v5-ui-polish-scrim-blur-behind-message-action-menu-shows-inner)

### 🎯 Goal

The scrim behind the message action menu showed a visible inner glow: a lighter ring hugging the left and right screen edges instead of an evenly dimmed background.

### 📝 Summary

- Blur the background snapshot with an opaque blur, so it no longer fades out along the screen edges
- Layer the scrim above the blurred snapshot instead of blurring it together with the snapshot

### 🛠 Implementation

`ReactionsOverlayView.backgroundView` composited the translucent scrim into the snapshot layer via `.overlay(...)` and then blurred the pair with `.blur(radius: 4)`.

SwiftUI's default non-opaque blur treats everything outside the layer's bounds as transparent, so the outer ~4pt of the composite got averaged toward clear. That thinned both the scrim and the snapshot at the edges and let the overlay's own `backgroundCoreElevation1` background bleed through as a lighter ring.

Two changes address it:

- `.blur(radius:opaque:)` with `opaque: true` clamps at the edges rather than fading to transparent.
- The scrim moved out of the blurred layer into a full screen sibling above it. Blurring a flat translucent colour has no visual effect other than thinning it at the edges, so it never belonged inside the blur.

As a side effect the scrim now spans the full screen rather than only the snapshot's frame, so on iPad with a visible tab bar the 20pt strip the snapshot is offset by is dimmed consistently with the rest of the screen.

### 🎨 Showcase

Edge pixel luminance measured on device screenshots of the open action menu, sampled across a uniform area of the scrim (iPhone 17 Pro, 3x, light mode):

| Pixel column from edge | 0 | 1 | 2 | 4 | 8 | 15 | centre |
| ---------------------- | - | - | - | - | - | -- | ------ |
| Before (left edge) | 196 | 192 | 188 | 181 | 168 | 152 | 141 |
| Before (right edge) | 195 | 192 | 188 | 181 | 169 | 152 | 141 |
| After (both edges) | 141 | 141 | 141 | 141 | 141 | 141 | 141 |

The ramp spanned roughly the 4pt blur radius at 3x. The centre value is unchanged, so the scrim keeps its density and only the edge falloff is gone.

### 🧪 Manual Testing Notes

Long press a message to open the action menu and look at the left and right screen edges: the background should be evenly dimmed, with no lighter band along the edges. Worth checking in both light and dark mode, and on iPad in both orientations (rotating while the overlay is open falls back to a plain scrim without the snapshot).

Note on unit tests: the existing `ReactionsOverlayView` snapshot tests pass unchanged, because the unit test process renders view layers without applying `CALayer` filters, so the blur never appears in recorded snapshots. This change is only observable in a real app process.

### ☑️ Contributor Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a visual issue on iOS where a lighter ring could appear around the screen edges of the blurred background behind the message action menu.
  * Improved background rendering during message reaction interactions for a cleaner, more consistent appearance.

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatMessageList/Reactions/ReactionsOverlayView.swift
+++ b/Sources/StreamChatSwiftUI/ChatMessageList/Reactions/ReactionsOverlayView.swift
@@ -208,18 +208,18 @@ public struct ReactionsOverlayView<Factory: ViewFactory>: View {
         ZStack {
             if !orientationChanged {
                 Image(uiImage: currentSnapshot)
-                    .overlay(
-                        Color(colors.backgroundCoreScrim)
-                            .edgesIgnoringSafeArea(.all)
-                            .opacity(popIn ? 1 : 0)
-                    )
-                    .blur(radius: popIn ? 4 : 0)
+                    // A non-opaque blur samples transparency outside the snapshot's bounds,
+                    // which fades its edges and lets the view's background bleed through as
+                    // a lighter ring. An opaque blur clamps at the edges instead.
+                    .blur(radius: popIn ? 4 : 0, opaque: true)
                     .edgesIgnoringSafeArea(.all)
                     .offset(y: overlayOffsetY)
-            } else {
-                Color(colors.backgroundCoreScrim)
-                    .edgesIgnoringSafeArea(.all)
             }
+            // The scrim is layered on top of the blur rather than blurred with the snapshot:
+            // blurring a translucent color only thins it out along the edges.
+            Color(colors.backgroundCoreScrim)
+                .opacity(orientationChanged || popIn ? 1 : 0)
+                .edgesIgnoringSafeArea(.all)
         }
         // The blurred snapshot and scrim are purely decorative; keep them out of
         // the accessibility tree so VoiceOver focuses the overlay's content.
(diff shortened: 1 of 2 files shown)


Sources/StreamChatSwiftUI/ChatMessageList/Reactions/ReactionsOverlayView.swift
```diff
@@ -208,18 +208,18 @@ public struct ReactionsOverlayView<Factory: ViewFactory>: View {
...
                     .offset(y: overlayOffsetY)
-            } else {
-                Color(colors.backgroundCoreScrim)
-                    .edgesIgnoringSafeArea(.all)
             }
+            // The scrim is layered on top of the blur rather than blurred with the snapshot:
+            // blurring a translucent color only thins it out along the edges.
+            Color(colors.backgroundCoreScrim)
+                .opacity(orientationChanged || popIn ? 1 : 0)
+                .edgesIgnoringSafeArea(.all)
```
nuno-vieira, line 222: I guess we should re-record the ReactionsOverlayView_Tests ?

nuno-vieira (approved):

Sources/StreamChatSwiftUI/ChatMessageList/Reactions/ReactionsOverlayView.swift
martinmitrevski: nope, locally it wasn't needed, let's see what the CI says

PR #1560, opened 2026-08-05
Apply dynamic search debouncing in channel list search

nuno-vieira:
### 🔗 Issue Links

- https://linear.app/stream/issue/IOS-1932/improve-search-debouncing-dynamic-debounce-based-on-query-length
- Depends on GetStream/stream-chat-swift#4198 (must merge first)

### 🎯 Goal

Search shouldn't send a network request on every keystroke, especially for very short text (1–2 letters), which is expensive for the backend and rarely useful. This PR adopts the same fix from the core SDK (stream-chat-swift) in SwiftUI: wait longer before searching for short text, and less time once the text is 3+ letters.

### 📝 Summary

- Channel search now uses the SDK's `ChatChannelSearchController`, which already waits before sending a request and cancels the previous request if you keep typing.
- Message search also relies on `ChatMessageSearchController` for the same reason. The view model now reuses one controller across searches instead of creating a new one each time, so cancelling actually works.
- The "Add Members" screen search now uses the SDK controller's built-in wait instead of a hand-rolled timer.
- Clearing the search field cancels any search that was still waiting to be sent, so it can't slip through and show results after you've cleared the text.
- Search results now keep updating live while a search is showing (e.g. if a channel in the results gets a new message).
- Demo app: added a setting to choose whether the channel list searches "Channels" or "Messages" — the same kind of option the UIKit demo app already has.
- Demo app: a "Multi Search" screen (channel + message + user search side by side) was built and tried out while working on this, then reverted for now. It's kept as a single commit in the history so it can be brought back later if it's useful.

### 🛠 Implementation

- `channelListSearchController` is deprecated. Channel search used to work by building a fresh `ChatChannelListController` for every search and swapping it in; now it goes through `ChatChannelSearchController` instead, which handles the request itself. If your code reads `channelListSearchController`, switch to `channelSearchController` — the old property still compiles, but it now always returns `nil`.
- The message search controller used to be recreated on every search, which meant the old one (with its in-flight request) was simply left behind instead of being cancelled. It's now created once and reused, so its own cancel-and-restart behaviour works as intended.
- Member search dropped its manual Combine `debounce(for:)` timer. It now waits the same amount as every other search in the SDK, controlled in one place instead of being duplicated here.
- The view model is now the delegate of both the main channel list and the search controllers. A small check makes sure a change coming from search results doesn't overwrite the main channel list, and the other way around.

### Notes

- This PR depends on the stream-chat-swift PR linked above. Right now, `Package.swift` and the Xcode project point at a fixed commit from that PR's branch (not the branch tip), so this PR can be reviewed and tested against a stable snapshot. This will be switched to a normal `develop` version once that PR merges.
- `channelListSearchController` returning `nil` instead of the real controller is a real behaviour change for any code still reading it directly, even though nothing fails to compile.

### 🧪 Manual Testing Notes

1. Channel list → search channels: type slowly. The first request should be noticeably delayed for a 1–2 letter query, then fire quickly once you've typed 3+ letters.
2. Same for the message search tab.
3. Type a full word quickly: only one request should go out, for the final text.
4. Clear the search field right after typing: no extra request should be sent.
5. Add Members screen: search should wait the same way before sending a request.
6. With search results open, have another device send a message to one of the matching channels: the result should update in place, without searching again.
7. Demo app → Configuration → "Channel List Search": switch between Channels and Messages, log back in, and confirm the channel list search now uses the chosen type.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search responsiveness with dynamic debounce timing while typing.
  * Prevented search results from replacing the main channel list.
  * Improved clearing of search results and state when search text is removed.
  * Added clearer handling for search errors and unavailable accounts.


Sources/StreamChatSwiftUI/ChatChannelList/ChatChannelListViewModel.swift
```diff
@@ -482,21 +493,35 @@ import UIKit
...
+                filter: .and([
+                    .autocomplete(.name, text: searchText),
+                    .containMembers(userIds: [userId])
+                ])
+            )
+            // Do not start watching any of the searched channels.
+            query.options = []
+            channelListSearchController?.delegate = nil
+            let searchController = chatClient.channelListController(query: query)
+            channelListSearchController = searchController
```
laevandus: Should we consider deprecating `channelListSearchController` and using the new controller instead? Otherwise our own code does not use it.

nuno-vieira: It is tricky, but I deprecated it, although it is kinda a behaviour-breaking change if someone is using this controller directly. Now it is unusable. But I guess for this feature it's not a big deal to break it. If everyone is OK to ship it like this, I'm fine as well

laevandus (approved):
✅ Verified that there is no flashing when searching (no spinner). Looks great

DemoAppSwiftUI/ChannelHeader/MultiSearchDemoView.swift
```diff
@@ -0,0 +1,543 @@
...
+        guard let cid = message.cid else { return "Message" }
+        return chatClient.channelController(for: cid).channel?.name ?? cid.id
+    }
+
+    private var messageText: String {
+        message.text.isEmpty ? "Attachment" : message.text
+    }
+}
+
+@MainActor
```
martinmitrevski: why do we add main actor on all the views? that's redundant

DemoAppSwiftUI/ChannelHeader/MultiSearchDemoView.swift
```diff
@@ -0,0 +1,543 @@
...
+// Copyright © 2026 Stream.io Inc. All rights reserved.
+//
+
+import StreamChat
+import StreamChatSwiftUI
+import SwiftUI
+
+/// Combined search prototype: channels, messages, and users in parallel using the state layer.
+@MainActor
+struct MultiSearchDemoView: View {
```
martinmitrevski: why do we add this? The scope of the ticket was the debouncing. While there's a need for multi search, it first needs some design work and Jurgen's input. For example, the messages icon doesn't look good:

Image

StreamChatSwiftUI.xcodeproj/project.pbxproj
```diff
@@ -1543,8 +1543,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/GetStream/stream-chat-swift.git";
 			requirement = {
-				kind = revision;
-				revision = b1842c5e95698c420b2462615ff9cee7d90dacb7;
+				branch = "add/improve-search-debouncing-dynamic-debounce-based-on-query";
```
martinmitrevski: reminder not to forget this

martinmitrevski (commented):
I don't think we need the multi search in this PR. We need to design and think that through properly.

DemoAppSwiftUI/ChannelHeader/MultiSearchDemoView.swift
nuno-vieira: I can revert it, but keep in mind, the search in state layer will be not possible to Manual QA

nuno-vieira: Reverted it for now

StreamChatSwiftUI.xcodeproj/project.pbxproj
nuno-vieira: Done

PR #1564, opened 2026-08-11
Fix quoted attachment preview size when thumbnail is unavailable

martinmitrevski:
### 🔗 Issue Links

[IOS-1896](https://linear.app/stream/issue/IOS-1896/swiftui-video-attachments-in-composer-on-quoting-message-are-bigger)

### 🎯 Goal

When quoting a message with a video attachment, the attachment preview inside the quote was sometimes much wider than its intended 40pt, pushing the author name and message preview into truncation.

### 📝 Summary

- Fix the attachment preview in quoted messages stretching when its thumbnail is not available.

### 🛠 Implementation

`MessageImagePreviewView` (also used by `MessageVideoPreviewView`) passed its `size` to `LazyLoadingImage` but did not constrain itself. `LazyLoadingImage` only applies a frame to the loaded image — its loading and failure states render a plain `Color`, which is fully flexible. Inside `ReferenceMessageView`'s `HStack` that placeholder competed with the `Spacer` and took up a large part of the quote.

This only reproduced for attachments whose thumbnail never resolved, which is why some videos were affected and others were not.

The preview now sets an explicit frame, so it keeps the same size in the loading, failure and loaded states. This applies to every surface that shows a quoted attachment preview: the message list, the composer and the edited message view.

### 🎨 Showcase

| Before | After |
| ------ | ----- |
| The preview takes up about half of the quote, and the author name and message preview are truncated ("Reply to Emm…" / "I took a short…"). | The preview stays a 40pt square, and the author name and message preview are fully visible. |

Covered by the `test_quotedMessageView_videoWithoutLoadedThumbnail` snapshot.

### 🧪 Manual Testing Notes

Quote a message with a video attachment whose thumbnail cannot be loaded (for example with the device offline, or a video without a usable `thumb_url`). The preview should stay a 40pt square and the author name and message preview should remain fully visible.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed oversized attachment previews in quoted messages when thumbnail images are unavailable.
  * Loading and failure placeholders now remain constrained to the intended thumbnail dimensions.
* **Tests**
  * Added coverage for quoted video messages with failed thumbnail loading.


nuno-vieira (approved):
LGTM! ✅

PR #1566, opened 2026-08-11
Fix unread banner flicker while channel is open

nuno-vieira:
### 🔗 Issue Links

- https://linear.app/stream/issue/IOS-1726/uikit-unread-banner-flickers-when-channel-is-open-and-receiving
- Depends on LLC: https://github.com/GetStream/stream-chat-swift/pull/4214

### 🎯 Goal

Prevent the SwiftUI jump-to-unread button from flashing when a channel is already open at the bottom and new messages arrive, without leaving the channel unread when the user leaves quickly after receiving a message.

### 📝 Summary

- Expose `canMarkRead` from `ChatChannelViewModel` to the message list
- Hide jump-to-unread while stuck to the bottom and already able to mark read
- Keep the existing `markRead` throttler (first request still runs immediately; later calls within the interval are coalesced)
- On disappear, cancel the throttler and call `markRead()` immediately when `shouldMarkChannelRead` is still true
- Temporarily point the StreamChat SPM dependency at the LLC `fix/unread-banner-flicker` branch

### 🛠 Implementation

SwiftUI was still showing jump-to-unread briefly for brand-new messages even when the user was at the bottom. The visibility guard (`showScrollToLatestButton || !canMarkRead`) prevents that flash while the user is stuck to the bottom and already able to mark read. Combined with optimistic `markRead` in the LLC branch, unread UI clears without removing the existing 3s throttler.

Leaving the channel cancels any pending throttled mark-read, then marks read immediately if `shouldMarkChannelRead` is true (at bottom, can mark read, first page loaded, not locally marked unread), so messages seen before leave are not left unread and no delayed mark-read can fire after leave.

### 🎨 Showcase

N/A — timing/flicker fix best verified in the simulator.

### 🧪 Manual Testing Notes

1. Resolve packages against the LLC `fix/unread-banner-flicker` branch
2. Open a channel and stay scrolled to the bottom
3. Have another participant send messages
4. Confirm the jump-to-unread button does not briefly appear
5. Scroll up, receive new messages, and confirm unread UI still appears as expected
6. Receive a message at the bottom, leave the channel immediately, and confirm the channel is marked read
7. Leave while scrolled up with unreads, and confirm the channel stays unread

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed flickering of the jump-to-unread control when new messages arrive in an open channel.
  * Improved unread-state handling when leaving a channel by marking messages as read when appropriate.
  * Kept the jump-to-unread control available when messages cannot yet be marked as read.

* **Tests**
  * Added coverage for read-state updates and throttling when leaving a channel.

* **Documentation**
  * Added the fix to the upcoming changelog.


laevandus (approved):
✅ Verified that the banner does not show up briefly.

PR #1573, opened 2026-08-17
Allow customizing channel info actions

martinmitrevski:
### 🔗 Issue Links

_Customer request: attach a system message to the channel when a user leaves a group, without forking `ChatChannelInfoView`._

### 🎯 Goal

Leaving a group could not be customized. `ChatChannelInfoViewModel.leaveConversationTapped` was `public` but not `open`, and the leave button plus its confirmation alert were private to `ChatChannelInfoView`. The only way to run custom logic when leaving (for example, sending a system message) was to fork the whole screen, which then drifts behind the SDK — a fork made four minor versions ago misses fixes such as the muted member icon added in 5.8.0.

### 📝 Summary

- `ChatChannelInfoViewModel.leaveConversationTapped(completion:)`, `blockUserTapped()` and `addUsersTapped(_:)` are now `open`
- `ChatChannelInfoViewModel.channelController` is now `public`, so subclasses can act on the channel
- New `ViewFactory.makeChannelInfoActionsView(options:)` slot for the actions section of the channel info screen
- New public `ChannelInfoActionsView` and `ChannelInfoActionsViewOptions`
- The leave and block confirmations are now driven by the view model's `leaveGroupAlertShown` / `blockUserAlertShown`, which were public but unused, and their copy is exposed as `leaveConversationConfirmation` / `blockUserConfirmation`

### 🛠 Implementation

The mute / block / leave section moved out of `ChatChannelInfoView` into a new public `ChannelInfoActionsView`, which also owns the block and leave confirmation alerts. Keeping the alerts in the parent would have limited a custom actions view to intercepting the button tap rather than the confirmation itself.

The two customization points cover different things and compose:

- **What an action does** → override the now-`open` methods on `ChatChannelInfoViewModel`. This is enough to send a system message before or after leaving, with no factory work and the default UI intact.
- **Which actions are shown and how they look** → override `makeChannelInfoActionsView(options:)`.

`ChannelInfoActionsViewOptions` carries the view model and a `leaveConversation` closure. The closure is a capability the slot needs rather than a second interception hook: it leaves the conversation *and* dismisses the info screen, popping the channel view too when the screen was pushed from the message list — behavior a replacement view cannot reproduce, since `notifyChannelDismiss()` is internal. It funnels through `leaveConversationTapped`, so a view model override applies to custom actions views as well.

The default implementation of the new slot lives in the `ViewFactory` extension, so existing custom factories keep compiling.

The confirmation copy moved to the view model as `ConfirmationPopup` values, matching how participant actions already describe their confirmations. That keeps the branching (group vs. direct message, block vs. unblock) unit-testable, lets integrators change the copy without replacing the view, and the alert presentation state lives on the two previously unused published properties, so the confirmation can also be triggered programmatically. The error alert now reuses `Alert.defaultErrorAlert` instead of building the same alert again.

Note that the leave action in the members sheet (`leaveGroupAction(onDismiss:onError:)`) still calls `removeMembers` directly and does not route through `leaveConversationTapped`; routing it would change the error semantics of an existing public API. That path remains customizable through the already-`open` `participantActions(for:)`.

### 🎨 Showcase

No visual changes. The extraction is pixel-identical: all existing channel info snapshots pass unchanged.

### 🧪 Testing Notes

Patch coverage is 84.6% (measured locally with `xccov` over the diff). What is left uncovered is SwiftUI closures that never run in this test target: the two button tap actions, the `.alert` content builders and the `Alert` construction. The unit test bundle is unhosted, so SwiftUI does not present alerts in it; everything those closures delegate to is covered instead. File coverage: `ChatChannelInfoViewModel` 87.0%, `ChatChannelInfoView` 88.6%, `ChannelInfoActionsView` 82.7%.

24 tests were added, covering the leave / delete paths and their error branches, the mute toggle, `leaveGroupAction`, `blockUserTapped`, the confirmations for group vs. direct message and block vs. unblock, the actions slot being used by `ChatChannelInfoView`, and the channel dismissal when leaving from the message list.

### 🧪 Manual Testing Notes

In the demo app, open a group channel → channel info, and check that mute, block (in DMs) and leave / delete conversation behave as before, including the confirmation alerts and the dismissal after leaving from both the channel list and the message list.

To exercise the customization points, subclass `ChatChannelInfoViewModel` and override `leaveConversationTapped` to send a message before calling `super`, or override `makeChannelInfoActionsView(options:)` in a custom `ViewFactory`.

### ☑️ Contributor Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a customizable channel-info actions view through the view factory.
  * Added support for customizing mute, block, leave, and delete-conversation actions.
  * Added configurable icons and confirmation prompts for leave and block actions.
  * Improved handling for group conversations and one-on-one direct messages.
* **Performance**
  * Improved scrolling performance by reducing repeated background-state reads.
* **Documentation**
  * Updated the upcoming changelog with the new customization options.
* **Tests**
  * Expanded coverage for channel-info actions, confirmations, dismissal behavior, and customization.


Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChannelInfoActionsView.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    public init(options: ChannelInfoActionsViewOptions) {
+        viewModel = options.viewModel
+        leaveConversation = options.leaveConversation
+    }
+
+    public var body: some View {
+        if viewModel.shouldShowMuteChannelButton
+            || viewModel.shouldShowBlockUserButton
+            || viewModel.shouldShowLeaveConversationButton {
```
nuno-vieira: We can probably merge these into one `viewModel.shouldShowInfoCard`, so that logic does not leak to the view and stays in the view model.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChannelInfoActionsView.swift
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+    }
+
+    private var leaveButton: some View {
+        Button {
+            viewModel.leaveGroupAlertShown = true
+        } label: {
+            HStack(spacing: tokens.spacingMd) {
+                Image(systemName: viewModel.showSingleMemberDMView ? "trash" : "rectangle.portrait.and.arrow.right")
```
nuno-vieira: Shouldn't this come from the appearance object?

Sources/StreamChatSwiftUI/ViewFactory/Options/ChannelViewFactoryOptions.swift
```diff
@@ -307,6 +307,28 @@ public final class MemberAddViewOptions: Sendable {
...
+
+/// Options for creating the actions section of the channel info screen.
+public final class ChannelInfoActionsViewOptions: Sendable {
+    /// The view model of the channel info screen.
+    public let viewModel: ChatChannelInfoViewModel
+    /// Leaves the group, or deletes the conversation in direct message channels, and dismisses
+    /// the channel info screen.
+    ///
+    /// Call it from the leave button of a custom actions view. To change what happens when leaving
+    /// the conversation, override `leaveConversationTapped` in ``ChatChannelInfoViewModel`` instead.
```
nuno-vieira: These feel a bit weird. Why have both options? Why don't we just keep the view model?

CHANGELOG.md
```diff
@@ -3,7 +3,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
+### ✅ Added
+- Add `ViewFactory.makeChannelInfoActionsView` for customizing the actions in the channel info screen [#1573](https://github.com/GetStream/stream-chat-swiftui/pull/1573)
+
 ### 🔄 Changed
+- Allow overriding the leave conversation, block user and add members actions in `ChatChannelInfoViewModel` [#1573](https://github.com/GetStream/stream-chat-swiftui/pull/1573)
```
nuno-vieira, line 8: This should also be in Added section, no need to put it in Changed 🤔

nuno-vieira (commented):
Overall LGTM, just added some few comments

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChannelInfoActionsView.swift
```diff
@@ -0,0 +1,113 @@
...
+        for confirmation: ConfirmationPopup,
+        action: @escaping @MainActor () -> Void
+    ) -> Alert {
+        Alert(
+            title: Text(confirmation.title),
+            message: confirmation.message.map { Text($0) },
+            primaryButton: .destructive(Text(confirmation.buttonTitle), action: action),
+            secondaryButton: .cancel()
+        )
+    }
```
laevandus: Inline it? Feels like it just adds unnecessary lines

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
```diff
@@ -80,7 +82,8 @@ import SwiftUI
         }
     }
 
-    var channelController: ChatChannelController!
+    /// The controller of the channel shown in the channel info screen.
+    public var channelController: ChatChannelController!
```
laevandus: Should it stay internal set?

nuno-vieira: It's in the PR description; it should be public so that customers can use it to do other actions if needed

Sources/StreamChatSwiftUI/ViewFactory/Options/ChannelViewFactoryOptions.swift
martinmitrevski: consistency with Android, plus it's easier this way. Otherwise, you need to re-implement the header and the whole channel info view.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChannelInfoActionsView.swift
martinmitrevski: Done in 68d7e25a — added `shouldShowActionsCard` to the view model (`shouldShowMuteChannelButton || shouldShowBlockUserButton || shouldShowLeaveConversationButton`) and the view just checks that. Covered by two tests.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChannelInfoActionsView.swift
martinmitrevski: Partly done in 68d7e25a. Moved the icon to the view model next to `leaveButtonTitle` as `leaveButtonIcon`, and the delete case now uses `images.trash` from the appearance object. There is no leave/exit icon in `Appearance.Images` (it lives in StreamChatCommonUI), so the group case still falls back to the `rectangle.portrait.and.arrow.right` symbol — the same icon name `leaveGroupAction` uses. Happy to add a `leaveChannel` entry in stream-chat-swift as a follow-up so both come from appearance. Note this re-recorded a few channel info snapshots: rendering a symbol via `Image(uiImage:)` instead of `Image(systemName:)` shifts antialiasing by ~200 px per image, visually identical.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChannelInfoActionsView.swift
martinmitrevski: Done in 68d7e25a — inlined into both `.alert` closures and removed the helper.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChatChannelInfoViewModel.swift
martinmitrevski: Good point — made it `public internal(set)` in 68d7e25a. Customers can read it (which is what the PR description is about) but cannot swap the controller from the outside. Tests still assign it through `@testable`.

CHANGELOG.md
martinmitrevski: Done in 68d7e25a — both entries are under Added now.

Sources/StreamChatSwiftUI/ViewFactory/Options/ChannelViewFactoryOptions.swift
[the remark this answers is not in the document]
martinmitrevski: Done in 68d7e25a — the docs now say "one-on-one direct message channels" in `ChatChannelInfoViewModel.leaveConversationTapped`, `ChannelInfoActionsViewOptions.leaveConversation`, `ChannelInfoActionsView` and `ViewFactory.makeChannelInfoActionsView`.

CHANGELOG.md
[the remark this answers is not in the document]
martinmitrevski: Not applying this one: every release section in CHANGELOG.md uses `# <version>` with `### <section>` (249 `###` headings, zero `##`). Changing only the Upcoming section would make it inconsistent with the rest of the file and with what the release tooling produces.

Sources/StreamChatSwiftUI/ChatChannel/ChannelInfo/ChannelInfoActionsView.swift
```diff
@@ -0,0 +1,124 @@
...
+        case blockUser, leaveConversation
+        var id: Self { self }
+    }
+
+    @State private var alertType: AlertType?
+
+    public init(options: ChannelInfoActionsViewOptions) {
+        viewModel = options.viewModel
+        leaveConversation = options.leaveConversation
+    }
```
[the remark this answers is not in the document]
martinmitrevski: Not applying this one: no public view or options type in the SDK is annotated `@MainActor` (`ChatChannelInfoView`, `MemberAddViewOptions`, `ThreadListHeaderViewOptions` and friends are not), and the target builds clean under Swift 6 strict concurrency without it — `body` is already main-actor isolated through `View`, and the options only store the view model reference and a `@MainActor` closure. Annotating the options would also stop integrators from constructing them off the main actor for no benefit.

nuno-vieira (approved):
LGTM! ✅

PR #1581, opened 2026-08-19
Fix clipped composer and voice play icons during layout animations

nuno-vieira:
### 🔗 Issue Links

- https://linear.app/stream/issue/IOS-1897/swiftui-the-image-inside-voice-message-play-icon-is-shifted-on-swipe

### 🎯 Goal

Keep circular button glyphs in place when a parent layout animation runs, instead of letting `clipShape` hide or shift the icon until the animation settles.

### 📝 Summary

- Isolate the voice recording play button geometry so swipe-to-reply no longer clips the play glyph
- Isolate the composer attachment picker button geometry so the plus icon stays inside the circle while the keyboard opens and closes

### 🛠 Implementation

On iOS 17+, both buttons now use `geometryGroup()`. That isolates the button’s layout from ancestor transforms, which is what was desyncing the circular clip from the SF Symbol / plus icon.

### 🧪 Manual Testing Notes

- Swipe a voice message to reply and release: the play icon should stay visible and centered in the circle for the whole spring-back
- Open and close the keyboard in a channel: the composer plus icon should stay inside the circular button

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- **New Features**
  - Added customizable channel information actions.
  - Added support for overriding channel information actions.

- **Bug Fixes**
  - Fixed the voice message play icon disappearing while swiping to reply.
  - Fixed attachment picker icon clipping during keyboard transitions.
  - Improved visual stability for playback and attachment controls during gestures and layout changes.


martinmitrevski (approved):
nice one ✅

PR #1584, opened 2026-08-20
Fix jumping back to latest messages after a mid-page jump

nuno-vieira:
### 🔗 Issue Links

https://linear.app/stream/issue/IOS-1980/swiftui-sending-message-after-jumping-to-mid-page-does-not-load-first

### 🎯 Goal

After jumping to a quoted message on a mid page, sending a new message (or tapping scroll to bottom) should load the first page and jump to the latest messages. UIKit already does this; SwiftUI did not.

### 📝 Summary

- Load the first page when the current user sends a message while not on the latest page
- Scroll to the newest message as soon as that first page arrives
- Use the same path for the scroll-to-bottom button, instead of waiting 0.5s after `loadFirstPage`

### 🛠 Implementation

`ChatChannelViewModel` now listens for `NewMessagePendingEvent`, matching UIKit. If the current user sent a channel message and the first page is not loaded, it loads the first page.

Both sending and tapping scroll-to-bottom set a flag and call `loadFirstPage()`. When the first page lands in `didUpdateMessages`, we scroll to the newest message and hide the scroll-to-bottom button.

A mid-page insert from the current user no longer scrolls to the newest message on the current page.

### 🎨 Showcase

N/A

### 🧪 Manual Testing Notes

1. Open a channel with enough messages to paginate
2. Quote a message on a mid page and tap it to jump there
3. Send a new message — the list should load the first page and jump to the new message
4. Repeat the jump, then tap the scroll-to-bottom button — the list should load the first page and jump to the latest messages without a noticeable delay

### ☑️ Contributor Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Sending a message after jumping to a quoted message now loads and displays the latest messages correctly.
  - The scroll-to-latest button now navigates immediately to the newest messages after jumping to the middle of a conversation.
  - New messages you send now trigger the appropriate latest-message loading behavior in conversations and threads.
  - Message navigation now correctly handles quoted replies and returns to the latest content when needed.

### DIFF
--- a/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
+++ b/Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
@@ -47,6 +47,7 @@ import SwiftUI
     private var loadingPreviousMessages: Bool = false
     private var loadingMessagesAround: Bool = false
     private var scrollsToUnreadAfterJumpToMessage = false
+    private var scrollsToLatestAfterLoadingFirstPage = false
     private var disableDateIndicator = false
     private var channelName = ""
     private var onlineIndicatorShown = false
@@ -299,13 +300,14 @@ import SwiftUI
     public func scrollToLastMessage() {
         if channelDataSource.hasLoadedAllNextMessages {
             updateScrolledIdToNewestMessage()
-        } else {
-            channelDataSource.loadFirstPage { [weak self] _ in
-                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-                    self?.scrolledId = self?.messages.first?.messageId
-                    self?.showScrollToLatestButton = false
-                }
-            }
+            showScrollToLatestButton = false
+            return
+        }
+
+        scrollsToLatestAfterLoadingFirstPage = true
+        channelDataSource.loadFirstPage { [weak self] _ in
+            guard let self, self.channelDataSource.hasLoadedAllNextMessages else { return }
+            self.finishScrollingToLatestAfterLoadingFirstPage()
         }
     }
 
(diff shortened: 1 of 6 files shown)


Sources/StreamChatSwiftUI/ChatChannel/ChatChannelViewModel.swift
```diff
@@ -295,13 +300,14 @@ import SwiftUI
...
-                    self?.scrolledId = self?.messages.first?.messageId
-                    self?.showScrollToLatestButton = false
-                }
-            }
+            showScrollToLatestButton = false
+            return
+        }
+
+        scrollsToLatestAfterLoadingFirstPage = true
+        channelDataSource.loadFirstPage { [weak self] _ in
```
martinmitrevski, line 308: hmm, are we loading the first page even when we loaded all the messages?

nuno-vieira: No, there's an early return if hasLoadedAllNextMessages is `true`

martinmitrevski: but that's in the completion handler after we loaded the page

nuno-vieira: No, before this function as well:

```swift
public func scrollToLastMessage() {
        if channelDataSource.hasLoadedAllNextMessages {
            updateScrolledIdToNewestMessage()
            showScrollToLatestButton = false
            return // <-----
        }

        scrollsToLatestAfterLoadingFirstPage = true
        channelDataSource.loadFirstPage { [weak self] _ in
```

martinmitrevski: 🤦  I thought the else is still there

martinmitrevski (approved):

PR #1590, opened 2026-09-04
Fix edit message not passing mentions from composer

nuno-vieira:
### 🔗 Issue Links

- [IOS-2005](https://linear.app/stream/issue/IOS-2005/edit-message-does-not-persist-mentions-swiftui-uikit)
- LLC: https://github.com/GetStream/stream-chat-swift/pull/4268 (merged)
- StreamChatTestTools compile fix: https://github.com/GetStream/stream-chat-swift/pull/4269

### 🎯 Goal

Editing a message in SwiftUI and adding `@` mentions does not persist those mentions. `MessageComposerViewModel` already collects mention state for new messages, but `edit()` only called `messageController.editMessage(text:attachments:)`.

This PR forwards the mention data collected during edit so edited messages keep user, role, group, `@here`, and `@channel` mentions.

### 📝 Summary

- `MessageComposerViewModel.edit()` now receives the mention state already collected in `sendMessage()`.
- That state is passed through to `messageController.editMessage` (`mentionedUserIds`, `mentionedRoles`, `mentionedGroupIds`, `mentionedHere`, `mentionedChannel`).
- StreamChat is pinned to `609ea69fb` so this builds against the merged mention APIs and the StreamChatTestTools `@retroactive` compile fix. Restore a version pin after the next LLC release.

### 🛠 Implementation

`sendMessage()` already runs `clearRemovedMentions()` and snapshots mention fields before deciding between edit and send. The edit path now takes those same values instead of dropping them.

### 🎨 Showcase

N/A — no visual change. Mentions persist after edit; the composer UI is unchanged.

### 🧪 Manual Testing Notes

1. Send a message without mentions, then edit it and add an `@user` mention. Confirm the mention is highlighted after save.
2. Edit a message that already has mentions and remove one `@user` from the text. Confirm that mention is dropped after save.
3. Edit a message and add `@here` / `@channel` (or role/group mentions where available). Confirm they persist after save.
4. Repeat in a thread reply.

### ☑️ Contributor Checklist

- [x] Changelog is updated with new localization keys (n/a - no new strings)
- [x] New code is covered by unit tests (n/a - wires existing LLC API; coverage is in stream-chat-swift#4268)

## Summary by CodeRabbit

- **Bug Fixes**
  - Editing a message now preserves all mentions, including mentioned users, roles, groups, `@here`, and `@channel`.
  - Mentions remain visible and functional after saving message edits.
  - Improved consistency when composing and updating messages with mention content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 # Upcoming
 
-### 🔄 Changed
+### 🐞 Fixed
+- Fix mentions not being persisted when editing a message [#1590](https://github.com/GetStream/stream-chat-swiftui/pull/1590)
 
 # [5.10.0](https://github.com/GetStream/stream-chat-swiftui/releases/tag/5.10.0)
 _August 27, 2026_
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(url: "https://github.com/GetStream/stream-chat-swift.git", revision: "609ea69fba757994c112724689e7a9b148835bbc")
+        .package(url: "https://github.com/GetStream/stream-chat-swift.git", revision: "789071c145519fc9423f298e6905db216864af51")
     ],
     targets: [
         .target(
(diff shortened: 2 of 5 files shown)


martinmitrevski (approved):
reminder about the branch dependency

